PR #5, opened 2019-08-26
expose client sent headers

sakari:
We expose only the headers specified in the oa spec. And to do that need to do
some hackery to lose all the other headers.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,7 +1,7 @@
 import * as server from './server';
 import * as assert from 'assert';
 import safe from '@smartlyio/safe-navigation';
-import axios from 'axios';
+import axios, { AxiosResponse } from 'axios';
 import * as runtime from './runtime';
 import * as FormData from 'form-data';
 
@@ -85,6 +85,21 @@ function axiosToJson(data: any) {
   return data;
 }
 
+function getContentType(response: AxiosResponse<any>) {
+  if (response.status === 204) {
+    return 'text/plain';
+  }
+  const type = response.headers['content-type'];
+  return type.split(';')[0].trim();
+}
+
+function getResponseData(response: AxiosResponse<any>) {
+  if (response.status === 204) {
+    return '';
+  }
+  return response.data;
+}
+
 export const axiosAdapter: ClientAdapter = async (
   arg: server.EndpointArg<any, any, any, any>
 ): Promise<any> => {
@@ -107,8 +122,8 @@ export const axiosAdapter: ClientAdapter = async (
   return {
     status: response.status,
     value: {
-      contentType: 'application/json',
-      value: response.data
+      contentType: getContentType(response),
+      value: getResponseData(response)
     }
   };
 };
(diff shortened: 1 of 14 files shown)


src/server.ts
```diff
@@ -90,31 +99,59 @@ function throwResponseValidationError(tag: string, originalValue: any, e: oar.Ma
...
+
+function voidify(value: {} | undefined | null) {
+  if (value && Object.keys(value).length > 0) {
     return value;
   }
   return null;
 }
 
+function cleanHeaders<H>(maker: oar.Maker<any, H>, headers: {} | null | undefined) {
+  const normalized = voidify(lowercaseObject(headers));
```
sakari: this here is a bit convoluted as we need to handle cases where there actually are specified headers and when no headers are specified in the open api spec

src/server.ts
```diff
@@ -90,31 +99,59 @@ function throwResponseValidationError(tag: string, originalValue: any, e: oar.Ma
...
+  const made = maker(normalized, {
```
sakari: when maker is Make<any, null> this fails as it does not know what to do with the headers it gets

src/client.ts
```diff
@@ -223,17 +228,24 @@ function fillInPathParams(params: { [key: string]: string }, path: string) {
...
-  return (ctx: ClientArg<any, any>) =>
+  return (ctx: ClientArg<any, any, any>) =>
     call({
       path: handler.path,
       servers: handler.servers,
       method: handler.method,
       params,
-      query: (ctx as { query?: any }).query,
-      body: (ctx as { body?: any }).body
+      headers: ((ctx || {}) as { headers?: any }).headers,
```
ganymedes01: ~maybe just being annoying, but I think `ctx ? ctx.headers : undefined` makes this easier for other people to read/understand.~ Actually, why not use your lib, `safe`, here instead?

src/server.ts
```diff
@@ -90,31 +99,56 @@ function throwResponseValidationError(tag: string, originalValue: any, e: oar.Ma
   return null as any;
 }
 
-function voidify(value: {} | undefined) {
-  if (value == null || Object.keys(value).length > 0) {
+function lowercaseObject(o?: {} | undefined | null) {
```
ganymedes01: Don't need `undefined` here as `o` is optional. Maybe remove the `null` as well. Why would `headers` ever be `null` here?

src/server.ts
```diff
@@ -90,31 +99,56 @@ function throwResponseValidationError(tag: string, originalValue: any, e: oar.Ma
...
+  if (o == null) {
```
ganymedes01: If `o` is undefined, this `if` won't catch it, and then as error will be thrown when `Object.keys(undefined)` gets run

sakari:
``` 
> null == undefined
< true
```

ganymedes01: oooh, I didn't notice this wasn't a `===` sign. nice!

PR #10, opened 2019-09-26
Upgrade TS version to v3.6.3

ganymedes01:


src/server.ts
```diff
@@ -17,17 +17,11 @@ export interface RequestBody<A> {
   value: A;
 }
 
-export interface Params {
-  [param: string]: unknown;
-}
+export type Params = object;
```
ganymedes01: This is pretty much the only change necessary.
TS 3.6 complains unless every object used as `Params` also have `[param: string]: unknown;`, and most of them don't since they have pre-defined fields instead.
Assuming your intention with that was to enforce `Params` (and `Headers`, and `Query`) were always an object, that behaviour should be maintained after this change

sakari: that might be ok, there was something fairly intricate about having unknown valued keys but if this works for us then it might be ok

sakari:
will this generate the same code? We rely on the code generation baked into typescript so if that changes we might break something. At least we might advertise the typescript version we require in peerdeps

ganymedes01:
> will this generate the same code?

Tested it and yes. Everything is the same except for the `ValueClass` declarations because now they only take two type arguments instead of three.

PR #13, opened 2019-10-03
Split

sakari:
split runtime to smaller pieces to make it nicer to work with. This also changes the library exports so that all the various runtime pieces are now under their own namespace

mainly this is a prelim work for to split the adapters and runtime from the oats generator part to to own libs

on top of https://github.com/smartlyio/oats/pull/10


ganymedes01:
I think jenkins may not be properly configured to run lint here. When I run it manually it complains, but the PR is still marked as green 🤔

src/axios-adapter.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function axiosToJson(data: any) {
+  if (data instanceof runtime.valueClass.ValueClass) {
+    return runtime.valueClass.toJSON(data);
+  }
+  return data;
+}
+
+export const bind: runtime.client.ClientAdapter = async (
```
ganymedes01: Maybe unnecessary, but having a base client (and server) adapter from which axios/koa can "inherit" would be nice to help extend oats to use different client/servers in the future (native node http? express?) 🤔

sakari: what would the base adapter do?

ganymedes01: Nothing. It'd be more of an interface sort of thing...
Seen as this only exports a single function, it may very well be unnecessary 😅

PR #15, opened 2019-10-25
Return value in err msg

ganymedes01:
As it is now, we have to look at the logs to figure out what the actual value is. This should save us some time.


sakari:
@ganymedes01 re: discussion about leaking possibly secrets in the error message and those being hard to remove. I think we could hint that the value is secret in the openapi spec itself and not put it to the error message.

PR #24, opened 2019-11-11
Use deifininte assignment for class properties

sund4e:
projects with `strictPropertyInitialization` fails for generated types because the properties are initialised inside the make helper. Using definite assignment tells TS that the property is initialised.


sakari:
nice, I have the *exact* same change in https://github.com/smartlyio/oats/pull/25 😁

sakari:
I merged this to my branch.

PR #25, opened 2019-11-11
Be strict

sakari:
projects running with strict: true are having hard time with the generated code. lets fix that.


hannu:
Awesome :+1: That was fast

PR #27, opened 2019-11-20
start of nock server adapter

sakari:
provide test server from oas spec using nock


sakari:
move code to new repo

PR #31, opened 2020-01-20
Support request context for handlers in server side

jlitola:
Adds support for being able to provide request context to endpoint
handlers. Introduces new type `EndpointWithContext<RequestContext>`,
and changes Koa adapter to allow extracting data from koa context as
request context for the OpenAPI endpoints.

### DIFF
--- a/adapters/koa/koa.ts
+++ b/adapters/koa/koa.ts
@@ -1,8 +1,11 @@
 import * as Router from 'koa-router';
-import * as Koa from 'koa';
+import { ParameterizedContext } from 'koa';
 import * as runtime from '@smartlyio/oats-runtime';
 
-function adapter(router: Router): runtime.server.ServerAdapter {
+function adapter<StateT, CustomT, RequestContext extends object>(
+  router: Router<StateT, CustomT>,
+  requestContextCreator: (ctx: ParameterizedContext<StateT, CustomT>) => RequestContext
+): runtime.server.ServerAdapter {
   return (
     path: string,
     op: string,
@@ -10,7 +13,7 @@ function adapter(router: Router): runtime.server.ServerAdapter {
     handler: runtime.server.SafeEndpoint
   ) => {
     const koaPath = path.replace(/{([^}]+)}/g, (m, param) => ':' + param);
-    (router as any)[method](koaPath, async (ctx: Koa.Context) => {
+    (router as any)[method](koaPath, async (ctx: ParameterizedContext<StateT, CustomT>) => {
       const files = (ctx as any).request.files;
       let fileFields = {};
       if (files) {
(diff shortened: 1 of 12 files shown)


runtime/src/server.ts
```diff
@@ -21,6 +21,8 @@ export type Headers = object;
 
 export type Query = object;
 
+export type RequestContext = any;
```
GSokol: Don't you want to make it more restricted?

jlitola: This is the widest possible type for the request context, from which the real request context needs to be extended. Due to maintaining backwards compatibility it needs to be pretty wide. The real request context type is defined in the project using the oats-koa-adapter.

runtime/src/server.ts
```diff
@@ -36,20 +38,31 @@ export interface EndpointArg<
...
+> = (ctx: ServerEndpointArg<H, P, Q, Body, RC>) => Promise<R>;
 
 export type SafeEndpoint = Endpoint<
   Headers | undefined,
   Params | undefined,
   Query | undefined,
   RequestBody<any> | undefined,
-  Response<number, any, any>
+  Response<number, any, any>,
+  RequestContext
```
GSokol: Why don't you incapsulate `Headers`, `Params`, and `Query` in the `RequestContext`?

jlitola: Because they in no way belong there. RequestContext can be used to provide request specific facilities to endpoint handlers, like logging & authorization. As an example, you can check Caesar PR https://github.com/smartlyio/caesar/pull/180 (work in progress).

adapters/koa/koa.ts
```diff
@@ -1,16 +1,19 @@
 import * as Router from 'koa-router';
-import * as Koa from 'koa';
+import { ParameterizedContext } from 'koa';
 import * as runtime from '@smartlyio/oats-runtime';
 
-function adapter(router: Router): runtime.server.ServerAdapter {
+function adapter<StateT, CustomT, RequestContext extends object>(
```
sakari: does requestcontext really need to extend object?

adapters/koa/koa.ts
```diff
@@ -29,18 +32,31 @@ function adapter(router: Router): runtime.server.ServerAdapter {
...
-  handler(adapter(router))(spec);
+/**
+ * Bind provided handlers for the OpenAPI routes
+ *
+ * Koa's default StateT and CustomT values are selected as defaults
+ * @param handler
+ * @param spec
+ * @param requestContextCreator
+ */
+export function bind<Spec, RequestContext = any, StateT = any, CustomT = {}>(
```
sakari: could these anys be unknowns or voids by default?

jlitola: RequestContext I could probably switch to void, but StateT, and CustomT types are what they are right now by the default in Koa. We can change them to voids, but then default return type will change from what it is right now.

adapters/koa/koa.ts
jlitola: Not really, can remove that.

Issue #40, opened 2020-03-02
provide type info to valueclasses


sakari:
as is we cannot type scalars like strings so its eg hard to do generic operations that would deal with only one *brand* of string like pick all urls from the data. As is there is no way to attach extra info to strings in javascript so to get this we would need to wrap the scalars to objects or something. This is kind of annoying though as unboxing the objects does not happen in all cases automatically. Also equality comparisons between boxed scalars is not nice.

If we would provide the type info in the containing valueclass object we could make the types at least accessible in runtime without changing the straight up js semantics.

PR #42, opened 2020-03-06
Include reflectable types

sakari:
add runtime type info for values so that we can type strings and generate values

todo

 - [x] include format to the string type

solves https://github.com/smartlyio/oats/issues/40

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -515,6 +515,19 @@ function generateValueClass(key: string, schema: oas.SchemaObject) {
         )
       ])
     ),
+    ts.createProperty(
+      undefined,
+      [
+        ts.createModifier(ts.SyntaxKind.PublicKeyword),
+        ts.createModifier(ts.SyntaxKind.StaticKeyword)
+      ],
+      ts.createIdentifier('reflection'),
+      undefined,
+      ts.createTypeReferenceNode(fromLib('reflection', 'NamedTypeDefinition'), [
+        ts.createTypeReferenceNode(ts.createIdentifier(oautil.typenamify(key)), [])
+      ]),
+      undefined
+    ),
     ts.createMethod(
       undefined,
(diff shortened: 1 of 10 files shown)


ganymedes01:
```ts
public static reflection: oar.reflection.NamedTypeDefinition<Item> = {
        name: "Item",
        definition: {
            type: "object",
            additionalProperties: false,
            properties: {
                "id": {
                    required: true,
                    value: {
                        type: "named",
                        reference: typeItemId
                    }
                },
                "name": {
                    required: true,
                    value: {
                        type: "string"
                    }
                }
            }
        },
        get maker() { return makeItem; }
    };
```

this works instead of `Object.assign` 😬

sakari:
yes. however its kinda annoying to do the "recursion" wrappers

src/generate-types.ts
```diff
@@ -675,6 +702,220 @@ function generateTopLevelClassBuilder(key: string, schema: oas.SchemaObject) {
...
+  schema: oas.SchemaObject | oas.ReferenceObject
+): ts.ObjectLiteralExpression {
+  if (oautil.isReferenceObject(schema)) {
+    const resolved = resolveRefToTypeName(schema.$ref);
+    const type: ts.Expression = resolved.qualified
+      ? ts.createPropertyAccess(resolved.qualified, 'type' + resolved.member)
+      : ts.createIdentifier('type' + resolved.member);
+    return ts.createObjectLiteral(
+      [
+        ts.createPropertyAssignment('type', ts.createStringLiteral('named')),
```
ganymedes01: Where are we putting the information about, for example, whether or not the property is a URL? I thought it would be under `type`?

sakari: at the moment this does not save the format field anywhere. 

Might be usable for test data generation purposes even though afaik oats does not validate the format either so :shrug:

sakari: I'll put the format to the type

PR #43, opened 2020-03-11
Type directed traversal

sakari:
use reflection types to figure out types of non-object values when doing map/pmap


ganymedes01:
Type inference seems to be a bit funny:
Added this to `common.yaml`
```
    Something:
      type: object
      required:
        - item
      properties:
        item:
          $ref: '#/components/schemas/Item'
```

Then tried:
```ts
import { Something, Item } from './tmp/common.types.generated';
import * as reflectionType from './runtime/src/reflection-type';

const item = Item.make({ id:'1', name:'toot' }).success();

const wtv = new reflectionType.Traversal(Something.reflection, Item.reflection).map(item, (leaf) => {
    
});
```

and `Root` and `Leaf` seem to have been resolved to `any`

runtime/test/generator.ts
```diff
@@ -0,0 +1,188 @@
...
+      return fc.anything();
+    }
+    if (type.type === 'object') {
+      return new GenObject(type.properties, type.additionalProperties);
+    }
+    return assert.fail('todo arb for missing type ' + type);
+  }
+}
+
+class GenObject extends fc.Arbitrary<any> {
```
ganymedes01: I was thinking that this generator might be useful outside of tests... I at least have had to build oats types by hand before 😬 I believe there's one of those In `v8` still

sakari: it cannot be in the distributed package because that would require fast-check which would just feel wrong. So somebody would need to take the time to create a new package. Hopefully in a new repo

sakari:
then nested package structure causes issues with linked packages not being enough when the linked package has an unpublished version https://github.com/yarnpkg/yarn/issues/2611

PR #45, opened 2020-03-13
Don't resolve installed modules

kazmiruk:
**Motivation**

I have `Module1` where I want to import openapi from `Module2`. Currently I can do it by relative path to `node_modules` and it works pretty fine. But after that I want to import openapi from `Module1` to `Module3`. And `Module2` installs as dependency of `Module2` in subfolder `node_modules`

```
/node_modules
    /Module1
        /node_modules
             /Module2
```

and when I try to build my `Module3` relative path starts to point on a wrong place in `node_modules` structure.
**My suggestion here**: allow using of installed modules names as root instead of relative path by setting `isInstalledModule` option. In this case module will not be resolved with relative path and we build types with imports from installed modules in `node_modules`.
Probably having additional flag is a little bit dirty approach but we can't rely only on node_modules in path

**Real example**

We have `v8` types which I want to use in `action library` (reuse Ad structure). And `action library` is used by `hermes` to build types.

### DIFF
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -23,6 +23,7 @@ function modulePath(importer: string, module: string | undefined) {
 export interface ImportDefinition {
   importFile: string;
   importAs: string;
+  isInstalledModule?: boolean;
 }
 
 export interface Driver {
@@ -58,7 +59,7 @@ export function generate(driver: Driver) {
       '\n' +
       types.run({
         externalOpenApiImports: (driver.externalOpenApiImports || []).map(i => ({
-          importFile: resolveModule(driver.generatedValueClassFile, i.importFile),
+          importFile: i.isInstalledModule ? i.importFile : resolveModule(driver.generatedValueClassFile, i.importFile),
           importAs: i.importAs
         })),
         externalOpenApiSpecs: driver.externalOpenApiSpecs || (() => undefined),
--- a/src/driver.ts
+++ b/src/driver.ts
@@ -59,7 +59,9 @@ export function generate(driver: Driver) {
       '\n' +
       types.run({
         externalOpenApiImports: (driver.externalOpenApiImports || []).map(i => ({
-          importFile: i.isInstalledModule ? i.importFile : resolveModule(driver.generatedValueClassFile, i.importFile),
+          importFile: i.isInstalledModule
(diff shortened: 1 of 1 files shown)


sakari:
nice. 

so you would like to have
```.ts
// action.types.generated.ts
import * as v8 from 'v8'
```
or something.

How are you going to pick the actual types from the imported package? Wouldn't you need to do have something like

```ts
import { types } from 'v8'
```

instead?

Looks good otherwise. I would kind of maybe have either a union type with two options like "import from file" and "import from package". Then telling what to import from the package would be a bit cleaner.  

Or then just assume that if the importFile parameter does not start with a '.' it' a package like normally with node instead of having a flag to do that.

kazmiruk:
> How are you going to pick the actual types from the imported package?

Currently I just use 

```
import * as v8 from 'v8';
const x: v8.AdCopy = ...;
```

> Or then just assume that if the importFile parameter does not start with a '.' it' a package like normally with node instead of having a flag to do that.

I thought about such approach but if library is public it could be inappropriate because currently we resolve module like relative path and we can break something by changing this logic.

cc @sakari

sakari:
I think that looking at the first character of the path is fine for determining whether its a module or a file path.

Not sure what you mean by the ´const x: ...´. If you are going to import the v8 package to the generated code how are you going to add the `const x: v8.AdCopy` thingy to there. Or are you going to actually import the thing to some other file and then export from there the types? If so that seems like a needlessly convoluted way to do this.

kazmiruk:
@sakari I meant that oas generates such imports as
```
import * as types from "./types/generated/oas-types.generated";
```

and replaces all required places with `types.<TypeName>`

Regarding resolving: if starts with `.` or `..` then resolve as relative path, if not - then as node module?

sakari:
referencing directly some file inside the package seems a bit meh as it requires some additional contract with the package instead of letting the package maintainer export the thing as they want to.

sakari:
re. resolving packages and '.'. yes

kazmiruk:
> referencing directly some file inside the package seems a bit meh as it requires some additional contract with the package instead of letting the package maintainer export the thing as they want to.

But how to export openapi file?

Issue #58, opened 2020-04-23
nullable schema objects are constructed incorrectly


sakari:
top level schema object with nullable property get constructed to an empty object when the value is null like from`{ field: null }` we get `{ field: Value {} }`

PR #59, opened 2020-04-23
Nullable bug

sakari:
split nullable classes to 'null | NonNullableSomeClass'
    
the class cannot be nullable as it needs to be a proper js class so we need to do some hackery. This fixes https://github.com/smartlyio/oats/issues/58 about constructing empty objects from nullable

The downside is that we end up with ugly NonNullableSomeClass for the actual class which also might break some existing code relying on `instanceof SomeClass`

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,9 @@ javascript class `NamedComponent`
  - For scalar types oats adds typescript branding to differentiate between various kinds of 
  named scalar types
  
+ For `type: object` schemas that are `nullable: true` the type is split to a `type NamedComponent = null | NonNullableNamedComponent` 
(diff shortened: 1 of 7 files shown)


ganymedes01:
Does it make sense, from a TypeScript perspective, to have nullable objects? In the end won't this behave the same as having a non-required property?

sakari:
```
null === undefined
false
JSON.stringify({ a: null})
"{"a":null}"
JSON.stringify({ a: undefined})
"{}"
```

there is a slight difference that causes odd behaviour when people are not careful with falsy values. eg. it is common, although kinda errorprone, to use null to *unset* a value when sending delta edits.

ganymedes01:
ahh... good point.

PR #74, opened 2020-07-01
Allowing consumers to ignore unsupported features

### LINKED ISSUE #54: Have a flag to ignore unsupported OpenAPI keys instead of blowing up
Yo! So we're trying to be sneaky and actually use some of the more advanced features of the OpenAPI spec, such as 'security'.

Now, you have been sneaky by asserting that such a thing is not there, and if it is, you break all the things.

Would it make sense to either configure oats so that you can explicitly allow and/or ignore unsupported keys, or even default to this behaviour?

YannickMeeus:
This isn't there yet, there's a bunch of duplication and verbosity that I'm going to try and break down a bit further.

I also have to add the test scenario we talked about earlier (I can't find the conversation but trust me, I think I remember what words were more or less used...)

Closes #54

### DIFF
--- a/examples/driver.ts
+++ b/examples/driver.ts
@@ -1,5 +1,6 @@
 // yarn ts-node examples/driver.ts
 import { driver, util } from '../index';
+import { UnsupportedFeatureBehaviour } from '../src/driver';
 
 // define how references to outside the example.yaml file are resolved
 const externals = {
--- a/examples/driver.ts
+++ b/examples/driver.ts
@@ -29,6 +29,20 @@ driver.generate({
   openapiFilePath: './test/example.yaml'
 });
 
+// generate server from the shared openapi spec
+// This example uses a specification file that contains compliant but unsupported nodes,
+// such as 'securitySchemes' and 'security'
+driver.generate({
+  ...externals,
+  generatedValueClassFile: './tmp/server.types.generated.ts',
+  generatedServerFile: './tmp/server.generated.ts',
+  header: '/* tslint:disable variable-name only-arrow-functions*/',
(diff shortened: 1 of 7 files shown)


YannickMeeus:
Test scenario added, the duplication stays until people give me their opinion on it, then I'll do one final pass.

examples/driver.ts
```diff
@@ -1,5 +1,6 @@
 // yarn ts-node examples/driver.ts
 import { driver, util } from '../index';
+import { UnsupportedFeatureBehaviour } from '../src/driver';
```
sakari, line 3: were you planning to use this for testing?

YannickMeeus: Obviously I forgot to push my latest local changes, which include the test...

sakari:
the api for configuring the thing seems workable so good to go from my pov. Were you planning to add to this something still?

YannickMeeus:
@sakari If you have any suggestions as to where to place that enum I introduced, that'd still be appreciated. It's currently duplicated in the driver *and* server and client generation modules, so if you think that could go into `util.ts`, I'd rather not have that live in all three modules 🙃

sakari:
It would be nice to (at least re)-export it from driver.ts. Other than that merging those multiple instances to some util.ts sounds cool to me. 

I'm a bit surprised that ts does not complain now about the enums being kinda nominally different

YannickMeeus:
@sakari Right, tidied some stuff up. Should be good for final review, merge, publish, all of that jazz.

I don't have le access to publish, me thinks, so I'll leave that in your hands.

PR #105, opened 2020-08-11
Support custom maker function using `x-oats-maker`

jlitola:
Allows one to use `x-oats-maker` to define custom maker function for
the type. This allows cleaning up, and sanitizing things when building
the data types.

E.g.

```
export const buildStringFromMongoId: Maker<any, ShapeOfMongoObjectId> = (value, opts) =>
  oar.make.makeString()(value.toString(), opts);
```

And in schema:

```
MongoObjectId:
  type: string
  x-oats-maker: 'mongoTools.buildStringFromMongoId'
```

You can use `externalOpenApiImports` to import functions into scope.

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -1042,21 +1042,25 @@ function generateTopLevelMaker(
       [
         ts.createVariableDeclaration(
           name + oautil.typenamify(key),
-          ts.createTypeReferenceNode(fromLib('make', 'Maker'), [
-            ts.createTypeReferenceNode(shape, []),
-            ts.createTypeReferenceNode(resultType, [])
-          ]),
-          makeCall(makerFun, [
-            ts.createFunctionExpression(
-              undefined,
-              undefined,
-              undefined,
-              undefined,
-              [],
-              undefined,
-              ts.createBlock([ts.createReturn(generateMakerExpression(schema))])
-            )
-          ])
+          schema['x-oats-maker']
+            ? undefined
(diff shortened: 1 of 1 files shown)


sakari:
ok. so what you are after here is *just* being able to define a serializers for doing `make({ id: someObjectId, ...}).success()` without having to say `make({ id: someObjectId.toString(), ...}).success()` right? 

You want the types emitted by the generator to be as in the schema. 

The like main issue I'm seeing here is that this kind of scheme would force everyone using oats and the schema you are providing to provide also the custom function -- I see many unhappy campers in the near future with that.

jlitola:
My original idea was to replace types altogether, but that seemed to be more involved change. I’m probably not able to take the time needed to implement that. For now it would be enough to sanitize the input, but it might be hard to move forward from this. We are not exposing the schema outside of v1 / caesar.

Anyway, probably best to ditch the approach, and I just deep-walk mongo response and convert ids to strings. Then we can use strings in schema without problems, as we are not writing back using oats generated types.

sakari:
would have been nice to get support for custom datatypes -- well, next time. 

I take it that toJSON for the object id instances does something else than just output a string?

jlitola:
Looks like it actually outputs just the string. Is oats calling toJSON in if it exists?

sakari:
no, I mean you can do JSON.parse(JSON.stringify(..)) 😬  to lose the extra bits unless you are keen to do traversals

jlitola:
Riiight. Anyway, closing the PR as this is not needed right now.

PR #123, opened 2020-09-05
Better driver

sakari:
having to define all the files used one by one is really tedious. Instead we can follow the dependencies as defined in the openapi spec and generate files as we go.

todo

 - [x] fix readme


src/driver.ts
```diff
@@ -45,38 +47,100 @@ function emitAllStatusCodes() {
...
+  };
 }
 
 export function generate(driver: Driver) {
   const file = fs.readFileSync(driver.openapiFilePath, 'utf8');
   const spec: oas.OpenAPIObject = yaml.load(file);
+
+  types.deprecated(
+    driver.generatedServerFile && driver.generatedClientFile,
+    'generating both server and client files from the same definition is frowned upon and will be prevented later on'
```
sakari, line 119: the client and server types might need to diverge when/if we get to support readonly/writeonly properties or @khugo s differing generated types for queries

src/generate-types.ts
```diff
@@ -5,230 +5,52 @@ import * as _ from 'lodash';
 import * as assert from 'assert';
 import * as oautil from './util';
 import { UnsupportedFeatureBehaviour } from './util';
```
sakari, line 7: I wrapped the whole function to a closure to get some asynchronicity safe state going on in `state` and `options` hence the big diff. The actual changes are in the top part of `run`

src/generate-types.ts
```diff
@@ -261,1117 +83,1396 @@ const statusCodes = [
...
+    actions: [] as Array<() => Promise<void>>
+  };
+  const builtins = generateBuiltins();
+  const types = generateComponents(options);
+  const queryTypes = generateQueryTypes(options);
+
+  const externals = generateExternals(
+    Object.entries(state.imports).map(([importAs, importFile]) => ({
+      importAs,
+      importFile: resolveModule(options.targetFile, importFile)
```
sakari, line 114: we resolve the paths that are given relative to cwd from resolve to paths relative to the import file to make things right.

src/generate-types.ts
```diff
@@ -261,1117 +83,1396 @@ const statusCodes = [
...
-  if (ref[0] === '#') {
-    return { member: oautil.refToTypeName(ref) };
   }
-  if (options.externalOpenApiSpecs) {
-    const name = options.externalOpenApiSpecs(ref);
-    if (name) {
-      const [qualified, member] = name.split('.');
-      return { member, qualified: ts.createIdentifier(qualified) };
+
+  function resolveRefToTypeName(ref: string): { qualified?: ts.Identifier; member: string } {
```
sakari, line 1441: this function does all the heavy lifting of applying the resolver

src/generate-types.ts
```diff
@@ -261,1117 +83,1396 @@ const statusCodes = [
...
+    });
+    return result.join('\n');
+  }
+
+  function addToImports(
+    importAs: string,
+    importFile: string | undefined,
+    action?: (() => Promise<void>) | undefined
+  ) {
+    if (!state.imports[importAs]) {
```
sakari, line 1425: this maintains the import list, actions to run state and the global state of which files we are already going to generate as we are bit loose on the promise waiting part.

jlitola:
This is amazing! Was just thinking that we would really need something like this.

src/driver.ts
```diff
@@ -39,44 +41,107 @@ export interface Driver {
...
-    return toModule;
-  }
+function defaultResolve() {
+  return undefined;
+}
 
-  const p = path.relative(path.dirname(fromModule), toModule);
-  if (p[0] === '.') {
-    return p;
+export function localResolve(ref: string) {
```
ganymedes01, line 55: I'm guessing  other than `local` the only remaining option to do a valid `$ref` would be the URL kind... if that's the case, why expose all this? Couldn't we just figure out what  type  the  `$ref` is and then  resolve it automatically?

sakari: not sure I follow. As a guess: you are asking why we expose `resolve` at all and not just do the right thing automatically?

as a default that would be nice -- however I did not want to change the behaviour of the resolution for existing users. 

In the greater scheme of things `resolve` allows also, in addition to fetching files from urls and more importantly defining *how* and where that happens, mapping names to custom implementations or shared implementations. 

For example we can do https://github.com/smartlyio/facebook-blueprint/pull/70/files#diff-ea1fc938c4289be40df8874165cbd861R3 which we cannot just "do-the-right-things"

ganymedes01: your guess it correct

> defining how and where that happens, mapping names to custom implementations or shared implementations

makes sense 👍

src/generate-types.ts
```diff
@@ -5,230 +5,48 @@ import * as _ from 'lodash';
...
-  return {
-    oneOf: contentTypeSchemas
+export interface Options {
+  forceGenerateTypes?: boolean;
+  header: string;
+  sourceFile: string;
+  externalOpenApiImports: readonly ImportDefinition[];
+  targetFile: string;
+  resolve: Resolve;
+  externalOpenApiSpecs?: (url: string) => string | undefined;
```
ganymedes01, line 31: This way the IDE also shouts at you for using it

```suggestion
  /** @deprecated Consider using 'resolve' instead */
  externalOpenApiSpecs?: (url: string) => string | undefined;
```

src/generate-types.ts
```diff
@@ -5,230 +5,48 @@ import * as _ from 'lodash';
...
-      return schema;
-    })
-  );
-  return {
-    oneOf: contentTypeSchemas
+export interface Options {
+  forceGenerateTypes?: boolean;
+  header: string;
+  sourceFile: string;
+  externalOpenApiImports: readonly ImportDefinition[];
```
ganymedes01, line 28:
```suggestion
  /** @deprecated Consider using 'resolve' instead */
  externalOpenApiImports: readonly ImportDefinition[];
```

src/generate-types.ts
```diff
@@ -261,1117 +79,1403 @@ const statusCodes = [
...
+  const externals = generateExternals(
+    Object.entries(state.imports).map(([importAs, importFile]) => ({
+      importAs,
+      importFile: resolveModule(options.targetFile, importFile)
+    }))
+  );
+  state.actions.forEach(action => action());
+
+  const sourceFile: ts.SourceFile = ts.createSourceFile(
+    'test.ts',
```
ganymedes01, line 126: what's `test.ts`?

sakari: it's the filename we lie we are writing to when generating the source text. It does not get written anywhere though. I imagine that it is used for error diagnostics or something but 🤷‍♂️

ganymedes01 (approved):
looks good to me 💯💫
this will make resolving files with loads of references much nicer

PR #131, opened 2020-09-09
do not make paths relative by force

sakari:
making paths relative breaks imports from node_modules. As this is all very
confusing add info logging to tell what we are doing.


ganymedes01 (approved):
lgtm

PR #142, opened 2020-10-09
Fix bool enums

kazmiruk:
I need type in smartly-actions with flag `enabled: boolean`. By this flag I want to choose what type to use: no additional fields for `false` and to required fields for `true`. Current oats implementation threats boolean enum as string. This PR contains fix

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -963,9 +963,9 @@ export function run(options: Options) {
               'enum',
               ts.createArrayLiteral(
                 schema.enum.map(i =>
-                  i === 'true'
+                  i === true
                     ? ts.createTrue()
-                    : i === 'false'
+                    : i === false
                     ? ts.createFalse()
                     : assert.fail('unknown enum ' + i)
                 )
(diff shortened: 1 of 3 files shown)


src/generate-types.ts
```diff
@@ -963,9 +963,9 @@ export function run(options: Options) {
               'enum',
               ts.createArrayLiteral(
                 schema.enum.map(i =>
-                  i === 'true'
```
sakari, line 966: there is probably some murky reason why this was like so that escapes me. Could you write a test in codegen.spec.ts that the runtime allows `typeGuard.maker(true).succes()" but not `typeGuard.maker(false).success()`?

this might break existing code but the breakage seems to happen in codegen already instead of runtime so thats kind of ok

sakari:
sweet

src/generate-types.ts
kazmiruk: @sakari done. If everything ok can you also describe how to deploy?

PR #148, opened 2020-10-30
fix optional query params throwing in server

### LINKED ISSUE #147: when there are only optional query parameters, not passing in any in the requests results in validation error
we convert the query to void when there are no query parameters which gets rejected by a type definition expecting an object :/ this is annoying, to say the least, but most people have worked around this by passing something in the query always.

One way to sort this out would be to generate different types depending on whether the type is used from client or server side. That would also allow supporting readonly and writeonly properties with would be sweet.

sakari:
fixes #147

mainly by using https://github.com/smartlyio/oats-runtime/pull/94

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@smartlyio/oats",
-  "version": "2.6.0",
+  "version": "2.7.0-beta.1",
   "license": "MIT",
   "description": "Openapi3 based generator for typescript aware servers and clients",
   "private": false,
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -8,9 +8,12 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "render": "yarn ts-node examples/driver.ts && yarn ts-node render.ts",
-    "clean": "rm -rf ./dist/* && rm -rf ./tmp",
-    "build": "yarn clean && yarn render && yarn jest && yarn lint && yarn tsc",
+    "generate:test": "find test -name driver.ts -exec yarn ts-node -r tsconfig-paths/register --project tsconfig.json {} \\;",
```
sakari, line 11: bit of code to find openapi files from under test/

PR #153, opened 2020-11-10
add header to client and server files

sakari:
this is nice when adding eslint ignores to the files


ganymedes01 (approved):
LGTM
although i'd argue that eslint should probably be setup to ignore these files instead. would probably speed linting up quite a lot considering how big these get 😬

sakari:
true

PR #161, opened 2020-12-07
Add reponse headers property

ganymedes01:
It would look something like this and would allow us to access the header properties that are declared in the openapi spec
```ts
{
    readonly status: 201;
    readonly value: {
        readonly contentType: "application/json";
        readonly value: common.ShapeOfItem;
    };
    readonly headers: {
        readonly location: string;
    };
}
```

https://github.com/smartlyio/oats-runtime/pull/106
https://github.com/smartlyio/oats-axios-adapter/pull/42

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -359,7 +359,7 @@ export function run(options: Options) {
         required.push(headerName);
         return { ...memo, [headerName]: headerObject };
       } else if (headerObject.schema) {
-        required.push(headerName);
+        if (headerObject.required) required.push(headerName);
         return { ...memo, [headerName]: headerObject.schema };
       }
       return memo;
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -439,7 +439,7 @@ export function run(options: Options) {
                     }
                   }
                 ),
-            headers: { type: 'object', nullable: true }
+            headers: { type: 'object' }
           },
           required: ['status', 'value', 'headers'],
           additionalProperties: false
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -438,14 +438,14 @@ export function run(options: Options) {
                       }
                     }
                   }
-                )
+                ),
+            headers: { type: 'object', nullable: true }
           },
-          required: ['status', 'value'],
+          required: ['status', 'value', 'headers'],
           additionalProperties: false
         };
         if (!oautil.isReferenceObject(response) && response.headers) {
           schema.properties!.headers = generateHeadersSchemaType(response.headers);
-          schema.required!.push('headers');
         }
         responseSchemas.push(schema);
       }
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -368,7 +368,9 @@ export function run(options: Options) {
       type: 'object',
       properties,
       required,
-      additionalProperties: true
+      additionalProperties: {
+        type: 'string'
+      }
     };
   }
 
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -352,6 +352,26 @@ export function run(options: Options) {
     };
   }
 
+  function generateHeadersSchemaType(headers: oas.HeadersObject): oas.SchemaObject {
+    const required: string[] = [];
+    const properties = Object.entries(headers).reduce((memo, [headerName, headerObject]) => {
+      if (oautil.isReferenceObject(headerObject)) {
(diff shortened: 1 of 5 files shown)


src/generate-types.ts
```diff
@@ -352,6 +352,26 @@ export function run(options: Options) {
...
+        required.push(headerName);
+        return { ...memo, [headerName]: headerObject.schema };
+      }
+      return memo;
+    }, {} as { [key: string]: oas.SchemaObject | oas.ReferenceObject });
+    return {
+      type: 'object',
+      properties,
+      required,
+      additionalProperties: true
```
ganymedes01, line 371: `true` so that we don't need to declare *all*  the headers

src/generate-types.ts
```diff
@@ -352,6 +352,26 @@ export function run(options: Options) {
...
+      } else if (headerObject.schema) {
+        required.push(headerName);
+        return { ...memo, [headerName]: headerObject.schema };
+      }
+      return memo;
+    }, {} as { [key: string]: oas.SchemaObject | oas.ReferenceObject });
+    return {
+      type: 'object',
+      properties,
+      required,
```
ganymedes01, line 370: should we make the declared headers `required`? 🤔

ganymedes01: The specification doesn't mention it: https://swagger.io/docs/specification/describing-responses/

sakari: There is a required flag you can give to the header parameter. If the thing is required then yes otherwise I think not.

https://swagger.io/docs/specification/describing-parameters/#header-parameters

src/generate-types.ts
```diff
@@ -423,6 +443,10 @@ export function run(options: Options) {
           required: ['status', 'value'],
           additionalProperties: false
         };
+        if (!oautil.isReferenceObject(response) && response.headers) {
+          schema.properties!.headers = generateHeadersSchemaType(response.headers);
+          schema.required!.push('headers');
```
sakari, line 448: so to check we would only need to provide a headers object in the response if there are response headers defined in the spec?

sakari:
you could add an test for this by copy pasting and changing test/optional-queries/* to suit this use case. The gist there is to make it somewhat easy to test various api knobs.

sakari:
tldr; sweet, was just wondering whether I should do this a few days ago

src/generate-types.ts
ganymedes01: Hm, what about `$ref` headers? is there an easy way to "resolve" the ref so we can check if that's required?

src/generate-types.ts
ganymedes01: I guess I could also check if there's actually any headers declared in there, so we don't end up having to set empty objects as `headers` 👍

src/generate-types.ts
sakari: Point. not really as such if we consider that the ref can be to an another file. Best we could do is emit the equivalent of `undefined | parameterType` for parameter components, but it would still require the property to be there. Unless we want to have a look on the runtime type checking code for undefined values.

making $ref optional might be ok or even making everything optional

sakari: Copoint. It actually might be that we inline parameter schemas. so this should actually work out of the box (mostly bc inlining would fail for out of file references)

yes. we do if you look eg at generateParameterType in generate-types.ts it uses deref which does this.

sakari (approved):
nice. with or without tests lgtm.

src/generate-types.ts
sakari: we drop extra headers for requests in server side would that work here or are we keen to allow also unknown headers?

sakari: additionalproperties are a bit me in general as those get turned into `{[key]: unknown}` which are not really nice to use.

sakari (changes requested):
maybe rethink a bit whether allowing additionalProperties for response headers is nice

sakari:
maybe typing the additional headers as strings would be nicer as that is pretty much what those are going to end up being

sakari (approved):

ganymedes01:
I agree that not allowing extra headers might be ideal.
how would we go about that? I guess it would be the adapter handling the request's job to drop the extra fields, but afaik it has no knowledge of which fields it should keep/omit

I guess it could change the oats-runtime to make using that `{ unknownField: 'drop' }` option, but I think that would require a bit of refactoring

PR #169, opened 2021-01-09
Use runtime shape of

sakari:


src/generate-types.ts
```diff
@@ -625,15 +625,11 @@ export function run(options: Options) {
     );
     const brand = ts.createExpressionWithTypeArguments(
       [
-        ts.createTypeReferenceNode(ts.createIdentifier('ShapeOf' + oautil.typenamify(key)), []),
+        ts.createTypeReferenceNode(ts.createIdentifier('never'), []),
```
ganymedes01, line 628: Do we still need to give the Shape then? Or are we keeping it for backwards compatibility reasons?

sakari: backwards compatibility. We already cause trouble with the branding change so lets spread the pain a bit.

ganymedes01 (approved):

PR #175, opened 2021-01-22
generate "isA" for scalars

ganymedes01:
## Scalars/Enums
```yml
something:
  type: string
  enum:
    - toot
    - doot
    - noot
```
```ts
Object.assign(typeSomething, {
    name: "Something",
    definition: {
        type: "string",
        enum: ["toot", "doot", "noot"]
    },
    maker: makeSomething,
    isA: (value: any) => makeSomething(value).isSuccess()
})
```

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -1310,6 +1310,29 @@ export function run(options: Options) {
     }
 
     if (isScalar(schema)) {
+      let isA: ts.ArrowFunction | undefined = undefined;
+      if (schema.enum) {
+        isA = ts.createArrowFunction(
+          undefined,
+          undefined,
+          [makeAnyProperty('value')],
+          undefined,
+          ts.createToken(ts.SyntaxKind.EqualsGreaterThanToken),
+          ts.createBinary(
+            ts.createCall(
+              ts.createPropertyAccess(
+                ts.createArrayLiteral(schema.enum.map(v => ts.createLiteral(v))),
+                'indexOf'
+              ),
+              undefined,
+              [ts.createIdentifier('value')]
+            ),
+            ts.createToken(ts.SyntaxKind.GreaterThanEqualsToken),
+            ts.createIdentifier('0')
+          )
+        );
+      }
+
       return [
         generateBrand(key),
         ts.createTypeAliasDeclaration(
@@ -1321,9 +1344,10 @@ export function run(options: Options) {
         ),
         generateTypeShape(key),
         generateTopLevelMaker(key, schema),
-        generateNamedTypeDefinitionAssignment(key, schema)
+        generateNamedTypeDefinitionAssignment(key, schema, isA)
       ];
     }
+
     return [
       ts.createTypeAliasDeclaration(
         undefined,
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -1321,28 +1321,15 @@ export function run(options: Options) {
         ),
         generateTypeShape(key),
         generateTopLevelMaker(key, schema),
-        generateNamedTypeDefinitionAssignment(
-          key,
-          schema,
-          ts.createArrowFunction(
-            undefined,
-            undefined,
-            [makeAnyProperty('value')],
-            undefined,
-            ts.createToken(ts.SyntaxKind.EqualsGreaterThanToken),
-            ts.createCall(
-              ts.createPropertyAccess(
-                ts.createCall(ts.createIdentifier('make' + oautil.typenamify(key)), undefined, [
-                  ts.createIdentifier('value')
-                ]),
-                'isSuccess'
-              ),
-              undefined,
-              []
-            )
-          )
-        )
+        generateNamedTypeDefinitionAssignment(key, schema, generateIsAForScalar(key))
       ];
+    }
+
+    let isA: ts.ArrowFunction | undefined = undefined;
+    if (schema.oneOf) {
+      isA = generateIsAForOneOf(schema.oneOf);
+    } else if (schema.items) {
+      isA = generateIsAForArray(schema.items);
     }
 
     return [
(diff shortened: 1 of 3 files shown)


src/generate-types.ts
```diff
@@ -1270,6 +1272,29 @@ export function run(options: Options) {
     }
 
     if (isScalar(schema)) {
+      let isA: ts.ArrowFunction | undefined = undefined;
+      if (schema.enum) {
```
sakari, line 1276: sweet

I wonder if we could do the same things for all scalars using the maker and asking for isSuccess?

it would be a bit of a lie though as 'deadbeef' may or may not be some identifier. However this is the case for enums also. Anyhow we could use the maker for enum also if we are more confident in our little lie with them.

ganymedes01: Changed it to all scalars:
```yml
    laughs:
      type: string
      pattern: (h[aeiou])+
```

should result in:
```ts
Object.assign(typeLaughs, {
    name: "Laughs",
    definition: {
        type: "string",
        pattern: "(h[aeiou])+"
    },
    maker: makeLaughs,
    isA: (value: any) => makeLaughs(value).isSuccess()
})
```

ganymedes01: Could we just use `isSuccess()` for everything? and ditch the `instanceof` stuff completely?
That way we'd all support `oneOf` types :D

ganymedes01: > it would be a bit of a lie though as 'deadbeef' may or may not be some identifier

What do you mean by identifier?

src/generate-types.ts
```diff
@@ -1294,10 +1304,134 @@ export function run(options: Options) {
...
+    assert(calls.length > 0, 'empty oneOf should not be possible');
+
+    return calls.length === 1
+      ? calls[0]
+      : calls
+          .slice(2)
+          .reduce(
+            (acc, memo) => ts.createBinary(acc, ts.SyntaxKind.BarBarToken, memo!),
+            ts.createBinary(calls[0]!, ts.SyntaxKind.BarBarToken, calls[1]!)
+          );
```
ganymedes01, line 1369: ugly but I couldn't find any way to create `(a || b || c ...)` for as many values as I wanted

src/generate-types.ts
sakari: say you have hexadecimal ids like UserIds and AccountIds. Whether 'deadbeef' is an account id or a user id depends on how it originated and that we cannot figure out as strings do not have identity

sakari:
I think with arrays we could subclass Array instead of inspecting the elements.

```
class Tagged extends Array<number> {}
const p: Tagged = new Tagged();
...
isA(a: any) { 
  return a instanceof Tagged
}
...
if (typeTagged.isA(p)) {
    console.log('aaa')
}
```

That would also allow us to enforce array constraints like minItems.

Other way to skin this would be to use makers but allow passing in an option to make it allow only ValueClasses where it otherwise would construct one.

src/generate-types.ts
sakari: not to say that having unscoped identifiers is that brilliant idea to start with.

ganymedes01:
> I think with arrays we could subclass Array instead of inspecting the elements.

Good point. Or we might also give false matches for empty arrays of different types.

ganymedes01:
> That would also allow us to enforce array constraints like minItems.

@sakari couldn't this be done in the `makeArray` method as well? Similar to the way we enforce string patterns

ganymedes01:
meh
```
> class Toot extends Array {}
> Array.isArray(Toot)
false

**Comment by sakari** [comment]
> couldn't this be done in the makeArray

sure, using a maker would be kind of fine (% making sure that the arrayish things really is correctly tagged thing -- although at the moment we do not enforce that either so 🤷 )

**Comment by sakari** [comment]
> since this extends Array

gah. And one cannot `extends ReadonlyArray`. Verrry annoying. Letting the thing be mutable seems like an accident waiting to happen.

**Comment by ganymedes01** [comment]
@sakari how about:
```ts
export type SomeArray2 = ReadonlyArray<Lol> & { private __oats_value_class_brand_tag: string; }
```

Then in `makeArray` we can set `__oats_value_class_brand_tag` to something unique, and in `isA` we check `Array.isArray(value) && value.__oats_value_class_brand_tag === 'something-unique'`

sakari:
you would have the tag there at runtime instead of keeping those fully phantom? Maybe it would do for arrays. Could use a symbol for the runtime tag though. I wonder how that behaves with object destructuring :/

Alternative would be to keep a reference to the object in some WeakMap but then we would break _.clone and thus safe-navigation would not work anymore.

src/generate-types.ts
```diff
@@ -149,6 +149,17 @@ export function run(options: Options) {
   const finished = addIndexSignatureIgnores(src);
   return finished;
 
+  function generateOatsBandProperty() {
```
sakari, line 152:
```suggestion
  function generateOatsBrandProperty() {
```

src/generate-types.ts
```diff
@@ -167,19 +178,7 @@ export function run(options: Options) {
...
-          ts.createToken(ts.SyntaxKind.PrivateKeyword),
-          ts.createToken(ts.SyntaxKind.ReadonlyKeyword)
-        ],
-        quotedProp(oatsBrandFieldName),
-        ts.createToken(ts.SyntaxKind.ExclamationToken),
-        ts.createKeywordTypeNode(ts.SyntaxKind.StringKeyword),
-        undefined
-      )
-    );
+    proptypes.push(generateOatsBandProperty());
```
sakari, line 181:
```suggestion
    proptypes.push(generateOatsBrandProperty());
```

sakari (approved):
So this *only* adds isA for scalars? If so good to go, maybe would be nice to have a test for the generated scalar isAs.

when releasing maybe would be good idea to --tag beta it first and test this in few projects

ganymedes01:
> So this only adds isA for scalars?

yes, what I really wanted was `isA` for `oneOf` but that will be tricky 😞

> when releasing maybe would be good idea to --tag beta it first and test this in few projects

will do 👍

PR #180, opened 2021-02-08
Add support for discriminator in oneOf

jlitola:
Generates separate makeOneOfWithDiscriminator if discriminator
configuration is specified for oneOf. This allows make to home into
correct maker immediately without trying all the options, which improves
performance a lot.

With Caesar actions the difference was:

```
Before

makeAction x 2,624 ops/sec ±1.85% (86 runs sampled)

After

makeAction x 342,554 ops/sec ±1.74% (85 runs sampled)
```

Requires https://github.com/smartlyio/oats-runtime/pull/127. Will need to
bump version numbers once that is in.

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -4,8 +4,10 @@ import safe from '@smartlyio/safe-navigation';
 import * as _ from 'lodash';
 import * as assert from 'assert';
 import * as oautil from './util';
-import { UnsupportedFeatureBehaviour } from './util';
+import { SchemaObject, UnsupportedFeatureBehaviour } from './util';
 import * as path from 'path';
+import { ReferenceObject } from 'openapi3-ts';
+
 const valueClassIndexSignatureKey = 'instanceIndexSignatureKey';
 
 interface ImportDefinition {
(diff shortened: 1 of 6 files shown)


src/generate-types.ts
```diff
@@ -791,6 +793,53 @@ export function run(options: Options) {
...
+        !schema.properties ||
+        !(discriminatorFieldSchema = schema.properties[discriminatorField])
+      ) {
+        throw new Error(`field ${discriminatorField} not defined, required by discriminator`);
+      }
+      if (!('type' in discriminatorFieldSchema) || discriminatorFieldSchema.type !== 'string') {
+        throw new Error('Only string type supported type supported');
+      }
+      if (!('enum' in discriminatorFieldSchema) || !discriminatorFieldSchema.enum) {
+        throw new Error('Must have enum');
```
ganymedes01, line 819: why do we need `enum`? 🤔

jlitola: If you don't supply manually created mappings, we try to extract the mappings from the types automatically. This works by going through the discriminator field in all the options, and picking up enum values from there.

ganymedes01: OHH right! cool 😁

jlitola: So with:

```
oneOf:
  - type: object
    properties: 
      type:
        type: string
        enum: [ value1 ]
  - type: object
    properties: 
      type:
        type: string
        enum: [ value2, value3 ]
discriminator:
  propertyName: type
```

We'll redirect value1 to first type, and value2 and value3 to second type.

sakari (approved):
sweet. I'll merge this and release as beta so people can test this.

PR #181, opened 2021-02-11
NPM Module based ref for using generated types from libraries

tigranbs:
I've stolen an idea from Andre to resolve npm modules and import already existing generated types, instead of referencing direct yml file and generating types all over again.

https://github.com/smartlyio/facebook-blueprint/pull/388/files

```
ExampleTypeSchema:
   $ref: "@smartly/some-package-with-oats-types#ExportedOatsTypeName"
```

This basically becomes something like this

```
export type ExampleTypeSchema = somePackageWithOatsTypes.ExportedOatsTypeName
```

The idea is to make it possible to directly re-use types instead of generating them for each project.
Having this logic inside Oats library itself will help to use this module resolver across projects like Goya, Mosby and Cesar to reference facebook-blueprint types for example.

### UPDATE

We have now sample implementation here https://github.com/smartlyio/goya/pull/2036 with Github workflow for validating generated openapi file instead of the original one.


src/driver.ts
```diff
@@ -176,6 +176,23 @@ export function generate(driver: Driver) {
   }
 }
 
+export function resolveModuleRef(dependencies: string[]): Resolve {
+  return (ref: string) => {
+    for (const dependency of dependencies) {
```
sakari, line 181: sneaky

sakari:
I see the practical benefit of this, *however* I'm somewhat certain that the $ref syntax you are proposing is not recognised by any other openapi tool or ide -- which makes putting that syntax to the spec files a bit nasty.

If that is really what you want then I think we could put this resolver to a new package for distribution as I'm a bit wary of promoting ways to break everyones tooling.

Alternative would be to have logic in the resolver that matches the $ref to a yaml file to some package exported type. If we have even a bit consistent naming conventions that could be doable. wdyt?

tigranbs:
> Alternative would be to have logic in the resolver that matches the $ref to a yaml file to some package exported type. If we have even a bit consistent naming conventions that could be doable. wdyt?

Good point, have some ideas how we can do this, trying out now with Goya -> V1 openapi transition with this kind of $ref implementation. Will update a PR when will have something that at least works as a prototype.
**Ideally** this shouldn't be braking change, rather just a nice addon, but I agree if we can't implement somewhat "smart" type resolver from specific yaml files, this $ref logic would be useless

sakari:
but above is ofc just a suggestion, I'm ok with going with this pr if we need to keep things moving, especially as the behaviour is opt in. Just thinking of @khugo s feeling here really 😄

khugo:
Thanks @sakari! This looks pretty cool, but indeed it would be good to ensure that we are not breaking e.g. the api documentation with this change. I'm not sure how that is currently working, as you certainly can't import from node_modules either in the api-documentation repo.

waltsu:
This was originally @lautis idea, but stealing it shamelessly: Should we differentiate the npm refernces from the remote-reference (https://swagger.io/docs/specification/using-ref/) by prefixing the path of the ref e.g. with `$ref: 'npm://@smartly..'`. Then it'd be clearer to the reader that this isn't standard behaviour.

waltsu:
> Thanks @sakari! This looks pretty cool, but indeed it would be good to ensure that we are not breaking e.g. the api documentation generation with this change. I'm not sure how that is currently working, as you certainly can't import from node_modules either in the api-documentation repo.

We'll need some preprocessor in CI that'll mangle the npm references away and builds standalone, standard openapi files that can be pushed to api-documentation repo. (See Tigran's comment https://smartlyio.slack.com/archives/C013ADWPF8T/p1613040345473100?thread_ts=1613038933.471500&cid=C013ADWPF8T)

sakari:
> npm://@smartly..'

That does not help much with tooling though -- like I have a working jump to definition in idea and would be nice if tooling would keep working without changes.

using the generated types from a package sound to me like very much an implementation detail -- not so much a spec level detail. If possible I would try to avoid mixing those to the spec.

like the end goal here is to have nicely shared type definitions in ts. One option could be to read this match from the package metadata (say package.json) of the imported modules in node_modules. Say if we put

```
  openapiSpec: { some name: 'import/path/for/module' }
```

to package.jsons? `some name` could be even a proper url to the canonical source of the api or an urn to satisfy the most rfc bound of us 😬

tigranbs:
After having this implementation https://github.com/smartlyio/goya/pull/2036 we now have here
```
TestFeedAutomation:
      $ref: 'npm://@smartly/blueprint-automation#FeedAutomation'
```
Where we copy `blueprint-automation/*.{yml,yaml}` into existing `openapi` folder and resolve `FeedAutomation` type path relative to `openapi` folder. This means that after building it with `generateNPMRelativePaths` function we get a self contained type folder which we can just copy to Documentation generation or to V1 and at the same time having a proper `npm` package references

**This becomes**
```
TestFeedAutomation:
      $ref: blueprint-automation/openapi.yml#/components/schemas/FeedAutomation
```

sakari:
@tigranbs could this be a separate package from oats -- the thing this does seems very specific to one kind of setup.

tigranbs:
@sakari that might be better I guess, to keep non-standard tooling separate than a regular OpenAPI specs ... will make a new repo like `oats-v2` ? would be happy to get some naming suggestions 😀

sakari:
oats-v2 might be according to our general antisocial guidelines 😁  -- but maybe oats-npm-resolver could do

PR #185, opened 2021-02-23
SENATORS-340 | Generate TS files with same openapi file structure

tigranbs:
Reproducing Oats file generation issue for nested directories with the same file names
https://smartlyio.slack.com/archives/C013ADWPF8T/p1614069594083100

**Update**

Added `preserveRefPathStructure` boolean into Driver options to give ability to keep directory structure during TS file generation from `$ref`, which handles case commented blow by converting directory name and openapi yml file name into module name which makes it unique.


sakari:
A solution here would be to generate the files mimicking the paths for the input files so that the generated directory tree would resemble the source directory tree.

Trouble here is that changing the paths to which the files are generated is not backwards compatible as people can and do import the recursively generated modules directly to their own code. Otoh, what we do now is not very nice as the behaviour is kind of broken when the filenames collide -- so would not like to leave the current way of doing things as the default behaviour.

Another concern might be cases where we should not generate the files to subdirectories due to some host project reasons. For that we would need to provide an option to keep doing what they are doing if that is what they want.

test/example.yaml
```diff
@@ -118,16 +118,7 @@ components:
...
-      additionalProperties: false
-      required:
-        - type
-      properties:
-        type:
-          type: string
-          enum: ['choice4']
-        field4:
-          type: string
+      $ref: './nested-test/example.yaml#/components/schemas/ChoiceItem45'
```
tigranbs, line 121: Now this becomes something like this
```
import * as NestedTestExample from "./nested-test/example.types.generated";
.....
.....
.....
export type ChoiceItem4 = NestedTestExample.ChoiceItem45;
```

examples/driver.ts
```diff
@@ -23,6 +23,7 @@ driver.generate({
   header: '/* tslint:disable variable-name only-arrow-functions*/',
   openapiFilePath: './test/example.yaml',
   resolve: driver.compose(driver.generateFile(), driver.localResolve),
+  preserveRefPathStructure: true,
```
tigranbs, line 26: The idea is to provide this boolean if we want to keep Input directory structure otherwise it should be default as we have already.

src/driver.ts
```diff
@@ -71,12 +72,16 @@ export function compose(...fns: types.Resolve[]): types.Resolve {
...
+  return [
+    parts[0],
+    ...parts
+      .slice(1)
+      .filter(str => str.length > 0)
+      .map(capitalize)
+  ].join('');
 }
 
 export function generateFile(): types.Resolve {
```
sakari, line 87: could we move preserve flag to here instead of threading it through driver?

src/driver.ts
```diff
@@ -87,9 +92,13 @@ export function generateFile(): types.Resolve {
     const localName = ref.replace(/^[^#]+/, '');
     const fileName = ref.replace(/#.+$/, '');
     const ymlFile = path.resolve(path.dirname(options.sourceFile), fileName);
-    const generatedFileName = `${path.basename(ymlFile).replace(/\.[^.]*$/, '')}.types.generated`;
-    const moduleName = makeModuleName(fileName);
+    const generatedFilePath = options.preserveRefPathStructure ? fileName : path.basename(ymlFile);
```
sakari, line 95: could we name the thing to just `preservePathStructure` ? That to me sounds understandable

sakari (approved):
looks legit -- did you test that this does the thing you wanted?

tigranbs:
> looks legit -- did you test that this does the thing you wanted?

Yes! Tried on V8 and facebook-blueprint it works fine 👍

PR #186, opened 2021-02-25
Optimize parsing speed by reducing reliance on exceptions

jlitola:
Update:

Now with 100% less breakage but same performance improvements. Preserve constructor, but add one internal option to sidestep build and use provided results.

Original message

createMakerWith tried creating new object with constructor and catching
the errors. When used inside oneOf it lead to constant throwing of
exceptions with isn't great for performance.

This commit introduces breaking change in that generated class
constructors are private, and only means of creating class is via make.
Make will use build function for creating the data, and only creating
new object if the built was successful. This allows us pass errors in
one of case without attempted creation of object and throwing the error.

With Goya benchmark following changes were observed.

Before:

```
makeAd with simple existing creative x 48,992 ops/sec ±1.61% (89 runs sampled)
makeAd with simple creative using warren image x 2,991 ops/sec ±1.75% (89 runs sampled)
```

After:

```
makeAd with simple existing creative x 184,005 ops/sec ±1.86% (89 runs sampled)
makeAd with simple creative using warren image x 34,668 ops/sec ±2.02% (88 runs sampled)
```

### DIFF
--- a/src/generate-types.ts
+++ b/src/generate-types.ts
@@ -634,55 +634,18 @@ export function run(options: Options) {
     );
   }
 
-  function generateClassConstructor(key: string) {
+  function generateClassConstructor() {
     return ts.createMethod(
       undefined,
-      [ts.createModifier(ts.SyntaxKind.PublicKeyword)],
+      [ts.createModifier(ts.SyntaxKind.PrivateKeyword)],
       undefined,
       'constructor',
       undefined,
       undefined,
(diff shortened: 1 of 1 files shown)


jlitola:
I'm open for doing this is less breaking fashion also if anyone has suggestions.

ganymedes01 (approved):
Looks go to me 👍

PR #199, opened 2021-04-13
Export generate-types as well

tigranbs:
For `oats-generator` project we need `Resolve` type from `generate-types` which is not exported now, so we use type `driver.Driver['resolve']` which is kinda OK-ish, but it's type also contains `undefined` which is not true in our use-case there, so we do ugly `!` at the end of function call to tell TypeScript that it should be always with a `Resolve` type.

So may we can export `generate-types` and use Resolve type directly?

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,4 +1,5 @@
 import * as driver from './src/driver';
 import * as util from './src/util';
+import * as types from './src/generate-types';
 
-export { driver, util };
+export { driver, util, types };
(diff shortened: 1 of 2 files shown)


sakari:
Could we export only `Resolve` and its dependencies and not the whole generate-types?

tigranbs:
@sakari done! Thought to have exported whole file just in case 🙂

sakari (approved):

PR #200, opened 2021-04-14
ShapeOf is no longer necessary

ganymedes01:
https://github.com/smartlyio/oats-runtime/pull/143
either `oats-runtime` or `oats` is going to be have to be merged with failing CI since this change breaks both 😅

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@smartlyio/oats",
-  "version": "2.15.1",
+  "version": "2.16.0",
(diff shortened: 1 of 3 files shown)


sakari:
is this a backwards compatible change?

ganymedes01:
@sakari no

sakari:
why do we want to lose it? I mean if this causes tons of work for others what do we provide in exchange?

sakari (approved):

Issue #211, opened 2021-06-02
Explicit https status code clashes with range definition in responses object


goalaleo:
When running the generator on a file that specifies a response object using both and explicit status code 403 and a range 4XX
```
      responses:
        403:
          $ref: "#/components/responses/403"
        4XX:
          $ref: "#/components/responses/4XX"
```

## Expected
oats to respect explicit code precedence over the range as mentioned in the openapi specification:

>If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.

source: https://swagger.io/specification/
under subheader "Responses object"

## Actual
oats throws an error `AssertionError [ERR_ASSERTION]: duplicate status code 403`

### Reproduction steps
git checkout the `smartly/update-openapi-specs` branch, and run the oats generator on `openapi/pena/openapi.yml`
Related PR: https://github.com/smartlyio/feeds-ui/pull/115

PR #213, opened 2021-06-17
allow wildcard overlap with literal status code

sakari:
see https://swagger.io/specification/

fixes https://github.com/smartlyio/oats/issues/211

### DIFF
--- a/src/status-codes.ts
+++ b/src/status-codes.ts
@@ -3,6 +3,10 @@ import * as assert from 'assert';
 export function resolvedStatusCodes(codes: string[]): Map<string, number[]> {
   const seen = new Set();
   codes.forEach(code => {
+    assert(
+      typeof code === 'string',
+      `HTTP status codes must be strings, got ${code} with type ${typeof code}`
+    );
     assert(!seen.has(code), `Duplicate status code ${code}`);
     assert(
       /^\dXX$/.test(code) || /^default$/.test(code) || /^\d\d\d$/.test(code),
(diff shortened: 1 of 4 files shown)


src/status-codes.ts
```diff
@@ -0,0 +1,69 @@
...
+    seen.add(code);
+  });
+  const record = new Map();
+  const matchers: RegExp[] = codes.sort().map(c => {
+    if (c === 'default') {
+      return /.*/;
+    }
+    return new RegExp(c.replace(/XX$/, '..'));
+  });
+  for (const code of statusCodes) {
```
ganymedes01, line 20: how would this behave if we have two matches? something like:
```yml
default:
  description: rest
404:
  description: not found
4XX:
  description: some error
```

all the `4XX` codes would also match `default` and might end up in either depending on the order in which they were declared in the file I guess 😣

sakari: https://github.com/smartlyio/oats/pull/213/files#diff-4919a218fd041747e2f69f1ec399a576035ec3018691faa95eefcc02a5f264c0R14 sorts the status codes to first have numbers and then letters so 400 gets sorted before 4XX which get sorted before default.

thats also the reason for the old school loop here to allow breaking from the loop when we get the first match

ganymedes01: oh nice! didn't notice the `sort()` there 👍

ganymedes01 (approved):
looks good 👌

sakari:
I'll add a better test to catch pin point the magic with sort still

PR #215, opened 2021-07-14
Add "success" status code, meaning 200 or 304

davissorenson:
We want to be able to match all successful status codes, which for now means 200 or 304 ("not modified"). Some of our internal services at Smartly.io implement 304, so it's nice to have a convenient way to do this.

### DIFF
--- a/src/status-codes.ts
+++ b/src/status-codes.ts
@@ -9,7 +9,10 @@ export function resolvedStatusCodes(codes: string[]): Map<string, number[]> {
     );
     assert(!seen.has(code), `Duplicate status code ${code}`);
     assert(
-      /^\dXX$/.test(code) || /^default$/.test(code) || /^\d\d\d$/.test(code),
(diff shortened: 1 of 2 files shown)


ganymedes01:
aren't other 2XX statuses also success?

davissorenson:
They are, I suppose we should include them for completeness but in practice I think they're very rare:

Screen Shot 2021-07-14 at 14 20 37

ganymedes01:
Also, having `success` isn't exactly conforming to the standard: https://swagger.io/specification/

ganymedes01:
see `Responses Object` ☝️

ganymedes01:
Screenshot 2021-07-14 at 14 23 18

So you could define `2XX` and `304` and have the same response for both

davissorenson:
Ok, I'll close this, and just reference the success response under both `2XX` and `304`.

PR #219, opened 2021-08-31
Add nock

sakari:
move oats-nock-adapter to oats monorepo


ganymedes01 (approved):
didnt realise this one was missing

PR #225, opened 2021-09-13
Parameterized makers

sakari:
use reflection types for runtime checking. why:

 - less code to run and generate as we do not need to generate the maker structures
 - feasible to do runtime optimisation
   - here we use the types to figure out better orders of object property checking making the case for type tagged unions nicer

This pr also loses the discriminator support because 1) its not standard following really (even if we would not agree with the standard 2) the runtime object checking optimization should first check the type tag (an enum with a single value) so its not necessary and 3) its a bit extra code. *Edit*  turns out discriminating required a *bunch* of extra logic as our test case has tons of nested and tagged oneOf groups and it just sucked without that extra logic. Well, now its there.

❗ I published the packages as `@2.18.3-alpha.17+db37bdb` if you want to test this


packages/oats-runtime/src/make.ts
```diff
@@ -533,6 +445,113 @@ export function makeNullable(maker: any) {
...
+    return 1;
+  }
+  return 0;
+}
+
+function fromObjectReflection(type: ObjectType): Maker<any, any> {
+  const comparisonOrder = Object.keys(type.properties).sort((aKey, bKey) => {
+    const a = type.properties[aKey]!;
+    const b = type.properties[bKey]!;
+    return priority(b) - priority(a);
```
sakari, line 499: higher priority gets tried first

packages/oats/src/generate-types.ts
```diff
@@ -1321,7 +1169,7 @@ export function run(options: Options) {
                 undefined,
                 [],
                 undefined,
-                ts.createBlock([ts.createReturn(generateMakerExpression(schema))])
+                ts.createBlock([ts.createReturn(generateReflectionMaker(key))])
```
ganymedes01, line 1172: Might be able to get rid of the `build*` fields entirely. I don't think they have a use outside the generated files.
We could move the `oar.make.createMaker(function () { return ...; })` so that `fromReflection` already returns that maker, and then tack that directly inside the `make` method of the classes. Something like this:

```ts
    static make(value: ShapeOfPhoto, opts?: oar.make.MakeOptions): oar.make.Make<Photo> {
      const make = oar.fromReflection(typePhoto.definition)(value, opts);
      return make.isError() ? oar.make.Make.error(make.errors) : oar.make.Make.ok(new Photo(make.success(), { unSafeSet: true }));
    }
```

packages/oats-runtime/src/make.ts
ganymedes01: a bit obscure that we use inverse priority here. easy to miss that we're sorting asc instead of desc
might be clearer if `0` was top priority, but this might be one of those topics like how to pronounce `gif` 🌝

packages/oats-runtime/src/make.ts
```diff
@@ -533,6 +444,113 @@ export function makeNullable(maker: any) {
...
+    type.type === 'boolean' ||
+    type.type === 'number' ||
+    type.type === 'integer'
+  ) {
+    return !!type.enum;
+  }
+  return false;
+}
+
+function priority(v: ObjectType['properties'][0]) {
```
ganymedes01, line 469: I believe this is the intended way of getting the value from a type 🌝
```suggestion
function priority(v: ObjectType['properties'][string]) {
```

packages/oats-runtime/src/make.ts
```diff
@@ -533,6 +444,113 @@ export function makeNullable(maker: any) {
...
+    if (isEnum(v.value)) {
+      // high chance this is a union type tag
+      if ((v.value as any).enum?.length === 1) {
+        return 4;
+      }
+      // enum matches less things than non enum
+      return 3;
+    }
+    if (v.value.type === 'string') {
+      // a pattern matches less things than a non pattern so lets try this first
```
ganymedes01, line 481: true but it might also be slower, specially if we are compiling a new regex every time

sakari: all the makers are cached on first run so we *should* not. I'm willing to be edumacated though

packages/oats-runtime/src/make.ts
```diff
@@ -533,6 +445,145 @@ export function makeNullable(maker: any) {
...
+      ? type.additionalProperties === true
+        ? makeAny()
+        : fromReflection(type.additionalProperties)
+      : undefined,
+    comparisonOrder
+  );
+}
+
+function fromUnionReflection(types: Type[]): Maker<any, any> {
+  const { discriminator, undiscriminated } = discriminateUnion(types);
```
sakari, line 523: this was not at all nice to write. Turns out that the previous discriminator code considered non singular enums as type tags which was a nice find.

packages/oats/src/generate-types.ts
sakari: getting rid of some of the code gen would be nice.

packages/oats-runtime/src/make.ts
sakari: de-invert'd

packages/oats-runtime/src/make.ts
ganymedes01: did this change the non-standard behaviour of our discriminator implementation? Meaning that before we accepted any discriminator value as long as it was unique per `oneOf` ref, but the OpenApi specification says that the value must contain the name of the schema

packages/oats-runtime/src/make.ts
ganymedes01: If we are *probably* caching the regexes, and doing this just to try to squeeze some performance gains out of making, I think this is fine 👍

packages/oats-runtime/src/make.ts
sakari: > OpenApi specification says that the value must contain the name of the schema

thats the non-standard compliant bit I was kinda annoyed -- also somewhat annoyed at the standard itself. 

This pr deprecates the whole explicit discriminator support (but I need to add the runtime code for that still as otherwise this becomes non backwards compatible which sucks a bit)

ganymedes01: > also somewhat annoyed at the standard itself.

Indeed. The way it was defined made little sense to me 😅

> This pr deprecates the whole explicit discriminator support

Oh, I thought this was you adding it back due to the degraded performance issues you mentioned

ganymedes01: > but I need to add the runtime code for that still as otherwise this becomes non backwards compatible which sucks a bit

sakari: added, but forgot to push 😬

ganymedes01 (approved):
LGTM 💫

packages/oats-runtime/src/make.ts
ganymedes01: took me a while to realise what this new code actually did. clever 👌

PR #234, opened 2021-09-27
Enforce min and max length with strings

ganymedes01:


jasujm (approved):

packages/oats-runtime/src/make.ts
```diff
@@ -172,6 +174,12 @@ export function makeString(
     if (typeof value !== 'string') {
       return getErrorWithValueMsg('expected a string', value);
     }
+    if (minLength != null && value.length < minLength) {
+      return getErrorWithValueMsg('expected a string with length greater than ' + minLength, value);
```
sakari, line 178:
```suggestion
      return getErrorWithValueMsg('expected a string with length of at least ' + minLength, value);
```

packages/oats-runtime/src/make.ts
```diff
@@ -172,6 +174,12 @@ export function makeString(
...
+    }
+    if (maxLength != null && value.length > maxLength) {
+      return getErrorWithValueMsg('expected a string with length smaller than ' + maxLength, value);
```
sakari, line 181: ditto .length === maxLength is allowed

ganymedes01: is it? it wasnt clear to me from the specs if it was inclusive or not

sakari: I would kinda read this https://cswr.github.io/JsonSchema/spec/basic_types/ as meaning that

PR #245, opened 2021-10-01
feat: implement `convert` option for transforming integer and number types from string to number

dmitrysteblyuk:
BREAKING CHANGE number and integer types are by default converted for path and query parameters

BREAKING CHANGE format option in a schema requires format to be registered with `runtime.make.registerFormat()`

BREAKING CHANGE integer type asserts value is an integer

BREAKING CHANGE change `runtime.make.makeNumber()` signature - first argument is `isInteger: boolean`

### DIFF
--- a/packages/oats-runtime/src/make.ts
+++ b/packages/oats-runtime/src/make.ts
@@ -7,7 +7,7 @@ import { discriminateUnion } from './union-discriminator';
 import { isEqual, uniq } from 'lodash';
 
 export class MakeError extends Error {
-  constructor(public readonly errors: ValidationError[]) {
+  constructor(readonly errors: ValidationError[]) {
     super(
       'tried to get success value from error: ' + errors.map(validationErrorPrinter).join('\n')
     );
@@ -63,32 +63,29 @@ export class Make<V> {
     this.errors = errors;
   }
 
-  public isError() {
+  isError() {
     return this.errors.length > 0;
   }
 
-  public isSuccess() {
+  isSuccess() {
     return !this.isError();
   }
 
-  public success(handler?: (e: this) => V): V {
-    if (!this.isSuccess()) {
-      if (handler) {
-        return handler(this);
-      }
+  success(handler?: (e: this) => V): V {
+    if (this.isSuccess()) {
+      return this.value as V;
     }
-    if (!this.isSuccess()) {
-      throw new MakeError(this.errors);
+    if (handler) {
+      return handler(this);
     }
-    assert(this.isSuccess());
-    return this.value as V;
+    throw new MakeError(this.errors);
   }
 
-  public group(groupMessage: string) {
+  group(groupMessage: string) {
     return new Make(null, [{ path: [], error: { groupMessage, errors: this.errors } }]);
   }
 
-  public errorPath(path: string) {
+  errorPath(path: string) {
     return new Make(
       null,
       this.errors.map(error => ({
@@ -98,7 +95,7 @@ export class Make<V> {
     );
   }
 
-  public map<R>(fn: (value: V) => R): Make<R> {
+  map<R>(fn: (value: V) => R): Make<R> {
     if (this.isError()) {
       return Make.error(this.errors);
     }
@@ -111,12 +108,15 @@ export class Make<V> {
 
 export interface MakeOptions {
   unknownField?: 'drop' | 'fail';
+  defaultConvert?: boolean;
 }
 
 export type Maker<Shape, V> = (value: Shape, opts?: MakeOptions) => Make<V>;
 
 function getErrorWithValueMsg<T>(msg: string, value: T): Make<T> {
-  return error(`${msg}, but got "${JSON.stringify(value)}" instead.`);
+  const stringified =
+    typeof value === 'object' || typeof value === 'string' ? JSON.stringify(value) : String(value);
+  return error(`${msg}, but got \`${stringified}\` instead.`);
 }
 
 function error<T>(error: string): Make<T> {
@@ -135,9 +135,7 @@ function getFormatter(format: string | undefined): Maker<any, undefined> {
     return () => Make.ok(undefined);
   }
   const formatter = formats[format];
-  if (!formatter) {
-    return () => Make.ok(undefined);
-  }
+  assert(formatter, `format "${format}" is not registered.`);
   return formatter;
 }
 
@@ -162,10 +160,10 @@ function getPatterner(pattern: string | undefined): Maker<any, void> {
 }
 
 export function makeString(
-  format: string | undefined = undefined,
-  pattern: string | undefined = undefined,
-  minLength: number | undefined = undefined,
-  maxLength: number | undefined = undefined
+  format?: string,
+  pattern?: string,
+  minLength?: number,
+  maxLength?: number
 ): Maker<any, string> {
   const formatter = getFormatter(format);
   const patterner = getPatterner(pattern);
@@ -198,10 +196,26 @@ export function makeString(
   };
 }
 
-export function makeNumber(min?: number, max?: number): Maker<number, number> {
-  return (value: any) => {
-    if (typeof value !== 'number') {
-      return getErrorWithValueMsg('expected a number', value);
+export function makeNumber(
+  isInteger: boolean,
+  min?: number,
+  max?: number,
+  convert?: boolean
+): Maker<unknown, number> {
+  return (x, { defaultConvert = false } = {}) => {
+    let value: number;
+    if (typeof x === 'number') {
+      value = x;
+    } else if (typeof x === 'string' && (convert ?? defaultConvert)) {
+      value = parseFloat(x);
+      if (!Number.isFinite(value)) {
+        return getErrorWithValueMsg('expected a number', x as any);
+      }
+    } else {
+      return getErrorWithValueMsg('expected a number', x as number);
+    }
+    if (isInteger && !Number.isInteger(value)) {
+      return getErrorWithValueMsg('expected an integer', x);
     }
     if (min != null && value < min) {
       return getErrorWithValueMsg('expected a number greater than ' + min, value);
@@ -392,10 +406,7 @@ interface FormDataArguments {
 export type Binary = File | Buffer | FormBinary;
 
 export class FormBinary {
-  constructor(
-    public readonly binary: Binary,
-    public readonly options?: FormDataArguments['options']
-  ) {}
+  constructor(readonly binary: Binary, readonly options?: FormDataArguments['options']) {}
 }
 
 enum FileBrand {}
@@ -404,11 +415,7 @@ export class File {
   // @ts-ignore
   private brand: FileBrand;
 
-  constructor(
-    public readonly path: string,
-    public readonly size: number,
-    public readonly name?: string
-  ) {}
+  constructor(readonly path: string, readonly size: number, readonly name?: string) {}
 }
 
 function checkBinary(value: any) {
@@ -559,10 +566,12 @@ export function fromReflection(type: Type): Maker<any, any> {
   if ((type as any).enum) {
     return makeEnum(...(type as any).enum);
   }
+  let isInteger = false;
   switch (type.type) {
     case 'integer':
+      isInteger = true;
     case 'number':
-      return makeNumber(type.minimum, type.maximum);
+      return makeNumber(isInteger, type.minimum, type.maximum, type.convert);
     case 'string':
       return makeString(type.format, type.pattern, type.minLength, type.maxLength);
     case 'boolean':
@@ -622,7 +631,9 @@ export function createMakerWith<Shape, Type>(
  * Throw error if there are duplicate koys
  * @deprecated
  */
-export function mergeMappings(...mappings: readonly { [key: string]: Maker<any, any> }[]): {
+export function mergeMappings(
+  ...mappings: readonly { [key: string]: Maker<any, any> }[]
+): {
   [key: string]: Maker<any, any>;
 } {
   return mappings.reduce((memo, current) => {
--- a/packages/oats-runtime/test/make-from-reflection.spec.ts
+++ b/packages/oats-runtime/test/make-from-reflection.spec.ts
@@ -253,13 +253,12 @@ describe('string', () => {
     );
   });
 
-  it('accepts if format is not defined', () => {
+  it('rejects if format is not defined', () => {
     const type: Type = {
       type: 'string',
       format: 'some-format'
     };
-    const fun = make.fromReflection(type);
-    fun('b').success();
+    expect(() => make.fromReflection(type)).toThrow('format "some-format" is not registered.');
   });
 
   it('rejects if format rejects', () => {
@@ -329,6 +328,22 @@ describe('number', () => {
     expect(fun(4).errors[0].error).toMatch('expected a number smaller than');
     expect(fun(3).isSuccess()).toBeTruthy();
   });
+  it('requires number to be integer', () => {
+    const fun = make.fromReflection({ type: 'integer', minimum: 1 });
+    expect(fun(1.5).errors[0].error).toMatch('expected an integer');
+    expect(fun(123).success()).toBe(123);
+  });
+  it('converts string to integer', () => {
+    const fun1 = make.fromReflection({ type: 'integer' });
+    expect(fun1('123', { defaultConvert: true }).success()).toBe(123);
+    expect(fun1('123').isError()).toBe(true);
+
+    const fun2 = make.fromReflection({ type: 'number', convert: false });
+    expect(fun2('12345', { defaultConvert: true }).isError()).toBe(true);
+
+    const fun3 = make.fromReflection({ type: 'number', convert: true });
+    expect(fun3('12345', { defaultConvert: false }).success()).toBe(12345);
+  });
 });
 
 describe('array', () => {
@@ -372,7 +387,7 @@ describe('object', () => {
         properties: {},
         additionalProperties: { type: 'unknown' }
       });
-      expect(fun([]).errors[0].error).toEqual('expected an object, but got "[]" instead.');
+      expect(fun([]).errors[0].error).toEqual('expected an object, but got `[]` instead.');
     });
 
     it('disallows constructor', () => {
--- a/packages/oats/src/generate-types.ts
+++ b/packages/oats/src/generate-types.ts
@@ -915,7 +915,7 @@ export function run(options: Options) {
         true
       );
     }
-    if (schema.type === 'number') {
+    if (schema.type === 'number' || schema.type === 'integer') {
       const enumValues = schema.enum
         ? [
             ts.createPropertyAssignment(
@@ -925,7 +925,7 @@ export function run(options: Options) {
           ]
         : [];
       const properties = [
-        ts.createPropertyAssignment('type', ts.createStringLiteral('number')),
+        ts.createPropertyAssignment('type', ts.createStringLiteral(schema.type)),
         ...enumValues
       ];
       if (schema.minimum != null) {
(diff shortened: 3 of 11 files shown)


dmitrysteblyuk:
Motivation - in Persona service we're moving most of our validation to open api spec. We need this additional feature as well as stricter rules for existing validation.

packages/oats-runtime/test/make-from-reflection.spec.ts
```diff
@@ -253,13 +253,12 @@ describe('string', () => {
     );
   });
 
-  it('accepts if format is not defined', () => {
```
ganymedes01, line 256: what’s the reasoning behind this? why throw errors?

sakari: I would kinda ask the same thing. format is kinda unregulated -- anything goes there. Throwing on unknown seems kinda onerous.

packages/oats/src/generate-types.ts
```diff
@@ -938,32 +938,15 @@ export function run(options: Options) {
...
-              ts.createArrayLiteral(schema.enum.map(i => ts.createNumericLiteral('' + i)))
-            )
-          ]
-        : [];
-      const properties = [
-        ts.createPropertyAssignment('type', ts.createStringLiteral('integer')),
-        ...enumValues
-      ];
-      if (schema.minimum != null) {
+      if (schema.convert != null) {
```
sakari, line 941: from where does this `convert` prop come from? I can't find it from the openapi 3.0 spec

packages/oats-runtime/src/make.ts
```diff
@@ -198,10 +196,26 @@ export function makeString(
   };
 }
 
-export function makeNumber(min?: number, max?: number): Maker<number, number> {
-  return (value: any) => {
-    if (typeof value !== 'number') {
-      return getErrorWithValueMsg('expected a number', value);
+export function makeNumber(
+  isInteger: boolean,
```
sakari, line 200: is this necessary? cant't we have different makers for reals and integers?

dmitrysteblyuk: you're right, i didn't know how to make it dry at first, but now i found the solution 👍

packages/oats/src/generate-types.ts
dmitrysteblyuk: removed! (P.S. there's actually a way to extend an open api options by using `x-` prefix, but i agree it's better not overuse it)
Anyways the only use case is parsing request parameters, and we want always to do this if required by schema type

dmitrysteblyuk:
i removed `convert` option and introduced maker options for parsing instead.

packages/oats-runtime/test/make-from-reflection.spec.ts
dmitrysteblyuk: we want to rely more on oats for request validation. So we're gonna use custom formats ("isoDate", "email", "resourceId" etc). Now without ensuring the format is defined we can easily lose validation (without knowing it!) after refactoring (e.g. it is as simple as registering a format after running `runtime.server.createHandlerFactory()`) or just by making a typo.

dmitrysteblyuk: It's not a big breaking change. So now if you have an unregistered format, `runtime.server.createHandlerFactory()` will throw (it is not nice admittedly bc the error is in runtime). So the options in this case is either to remove `format` from schema (it is quite useless without validator - can be moved to description or summary) or registering it. I'll add more informative exception.

dmitrysteblyuk:
I forgot, i'll need to reflect these changes in client as well.

packages/oats-runtime/src/make.ts
```diff
@@ -246,15 +272,20 @@ export function makeOptional(maker: any) {
...
+  return (x: unknown, { parseBooleanStrings = false } = {}) => {
+    let value: boolean;
+    if (typeof x === 'boolean') {
+      value = x;
+    } else if (parseBooleanStrings && x === 'true') {
+      value = true;
+    } else if (parseBooleanStrings && x === 'false') {
+      value = false;
+    } else {
+      return getErrorWithValueMsg('expected a boolean', x as boolean);
```
taha-k, line 285: does boolean check have to be this specific? Ideally I would prefer anything that is not `false | null | undefined` be evaluated as `true`

dmitrysteblyuk: so `"gibberish"` will be considered as `true`?

dmitrysteblyuk: the original implementation required value to be of boolean type, so it didn't allow null, undefined etc.

packages/oats-runtime/test/make-from-reflection.spec.ts
sakari: It kind of does break things and I'm not convinced that requiring format is important enough to offset the cost.

Openapi 3 is way more lax on its handling of format

> Tools that do not recognize a specific format MAY default back to the type alone, as if the format is not specified

bunch of the string validation cases can be dealt with by `pattern`, would those work for you instead of having to define formats?

dmitrysteblyuk: some of our formats are more complicated than just using regex (and regexp for things like email/isoDate is also not that trivial). Besides this functionality `registerFormat()` was added so we can use it (and we want to!).
I'm wondering, what's the use case of having a format if there's no handler for it?
If it is really breaking, we can introduce some `assertFormatsRegistered` option somewhere (any ideas where?) to disable this assertion. We can even default it to `false`.

packages/oats-runtime/src/make.ts
```diff
@@ -111,12 +108,16 @@ export class Make<V> {
 
 export interface MakeOptions {
   unknownField?: 'drop' | 'fail';
+  parseNumberStrings?: boolean;
```
sakari, line 111: these seem much nicer approach. I'm wondering if we would want to have a parsers for any value here instead of just for numbers and strings. Going with this might be ok but I'm kinda worried that we end up with an api that we want to change later leading to breakage. Maybe I worry too much though 🤷 

@ganymedes01 wdyt?

packages/oats-runtime/src/make.ts
sakari: if we would map `undefined` to `false` how would optional boolean values be handled? To me not providing a value is different to setting it to `false`.

I other words I would go with the strictest possible interpretation

packages/oats-runtime/src/make.ts
dmitrysteblyuk: the only place where we want to have parsing is parameters (query/path) which are strings. We don't want to have parsing for body, bc you can have any type you want there. Now what we want to parse besides numbers and booleans? Idt nulls or undefineds are good ideas.

packages/oats-runtime/src/make.ts
taha-k: Its just an opinion, but I strongly feel that the result should be evaluated by the expression `Boolean(value)`. So `Boolean('gibberish')` will be true and even `Boolean('false')` will also be true. If one really wants a false, then the value can be anything that evaluates to false like `'' | false | null | 0 `

dmitrysteblyuk: I'd go for strictness too. Most times, following strict rules is easy, debugging loosely written code is hard.

taha-k: @sakari a boolean `undefined` does evaluate to `false`, I don't see any other way an optional boolean undefined should be interpreted. I do agree with strictness when its necessary but its also a balance that code does not end up being rigid to changes. However, I am unaware of the full context, so I concede to what you guys feel is more feasible
```
if (shouldDoSomething && shouldDoSomething === true) {
  // do something
} else {
  // don't do something
}
```

sakari: not having a property is different to having a property eg in cases where you try to transmit a delta with some homegrown framework (ofc a questionable idea in general)

packages/oats-runtime/src/make.ts
sakari: mmm. yes minimum amount of magic is preferable. Maybe we are fine with just numbers and strings. If not we can just add more options

packages/oats-runtime/test/make-from-reflection.spec.ts
sakari: could we have the format assertion as a separate discussion (and pr)? I think we, or at least I, agree on having the number and string parsers added to the query runtime checking so moving that forward is likely something that would benefit you even without the formatters.

packages/oats-runtime/src/make.ts
ganymedes01: We might want to parse arrays also, which don't even have a standard implementation (or do but no one follows it). Currently we define it as a string and then `split(',')`

packages/oats-runtime/test/make-from-reflection.spec.ts
dmitrysteblyuk: i thought we agreed on a driver option, but i can add in the next pr, sure.

packages/oats-runtime/src/make.ts
ganymedes01: Might be worth noting that OpenAPI spcs are not a JavaScript thing, so using JavaScript concepts to determine what is `false` and `true` might lead to APIs implemented in other languages behaving differently when their clients are implemented in JavaScript (or vice versa)

Not every language agrees that `''` is falsey, but every language agrees that `false` is `false`.

packages/oats-runtime/src/make.ts
dmitrysteblyuk: i'll look into this. This pr is still WIP bc i forgot to reflect its changes in the client.

ganymedes01: I agree that this is a much better solution than the one before 👍 But I also think having something that works for any values (as sakari suggested) might be better, otherwise we could end up having to grow this list of options every time we find some exception

dmitrysteblyuk: Afaiu open api primitive data types are basically JSON data types except `null`.
So these are - `number`, `boolean`, `string`. And we got it all covered i think.

ganymedes01: but whos to say that query parameters will only be primitives?

dmitrysteblyuk: >We might want to parse arrays also, which don't even have a standard implementation (or do but no one follows it). Currently we define it as a string and then split(',')

i was thinking why do we need this if it's possible to provide an array in query parameters?
koa (and others like express) are automatically creating an array if query param repeats itself.
What we can do in this case is too always convert the string (or number/boolean) to an array if it consists only of one element.
e.g. `?someItems=Item1&someItems=Item2` => will be handled by koa: `someItems: ['Item1', 'Item2']`
`?someItems=Item0` => not handled by koa: `someItems: 'Item0'`

dmitrysteblyuk: but now i'm facing this really weird issue.
Axios is appending `[]` to query param if it's an array (wtf 😅).
Now this complicates the matter. If anyone knows how to disable this, i'd be really grateful.

ganymedes01: having repeated parameters suffixed with [] is one of the many ways to represent arrays in query parameters

dmitrysteblyuk: yes, but it's a convention, not a requirement. For some reason it's being forced when you use axios.
https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js#L42
So if we want to support array schema for query parameters, the parameter names must be suffixed with `[]` (in openai spec), bc axios does it automatically (whether we want this or not). Now there's a way to provide `paramsSerializer`, but if we want to do we have to implement it.

packages/oats-axios-adapter/index.ts
```diff
@@ -42,7 +42,10 @@ export function withAxios(axiosInstance: AxiosInstance): runtime.client.ClientAd
       return assert.fail('cannot decide which server to use from ' + arg.servers.join(', '));
     }
     const server = arg.servers[0];
-    const params = axiosToJson(arg.query);
+    // Create `URLSearchParams` instead of passing raw data to axios.
+    // Otherwise axios will suffix array query parameters with "[]".
+    // See https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js#L42
+    const params = createURLSearchParams(axiosToJson(arg.query));
```
dmitrysteblyuk, line 48: now this is all not ideal. I'm open to suggestions. Also `URLSearchParams` is not supported by IE.

dmitrysteblyuk: alternative solution: https://github.com/axios/axios/issues/2840#issuecomment-602992060

packages/oats-runtime/src/make.ts
sakari: how the query parameters are mapped from and to a non scalar input is defined by the openapi spec. If we really want to we can work to support that bit better but that seems like lots of work and careful reading of the spec would be required as well as dealing with the various weird ways the adapters behave.

I would not do that now as I assume you have bunch of other stuff to do also.

packages/oats-runtime/src/make.ts
```diff
@@ -284,10 +295,19 @@ export function makeBoolean(): Maker<unknown, boolean> {
...
+export function makeArray(
+  maker: any,
+  minSize?: number,
+  maxSize?: number
+): Maker<unknown, unknown[]> {
+  return (x, opts) => {
+    let value: unknown[];
+    if (Array.isArray(x)) {
+      value = x;
+    } else if (opts?.allowConvertForArrayType) {
```
sakari, line 307: why we would want to do this?

dmitrysteblyuk: to support array schemas for query parameters
`param=a&param=b` is converted by koa to `query: {param: ['a', 'b']}`
but
`param=a` is converted to `query: {param: 'a'}`
So if we have an array schema for a query parameter, it'll fail unless we tell it to convert a single element to an array

packages/oats-runtime/src/make.ts
dmitrysteblyuk: > I would not do that now as I assume you have bunch of other stuff to do also
yeah i've already done it. It shouldn't break much if axios is used to make requests to server that uses oats. Since currently making array schemas for query params work with oats is quite complicated, it's unlikely anyone uses it. But i can check in our services.
Now it should be working quite nicely.

dmitrysteblyuk: i can move it to another pr of course. It'll mean 2 major versions in a row (most likely), but i guess it's fine.

packages/oats-runtime/src/make.ts
dmitrysteblyuk: interesting... Idt we're supporting any other style but `form` as of now. So the fact that we're adding support for repeated query params and arrays (still for the same `form` style) isn't breaking anything in our code or against open api spec.
The only thing that i may be missing is this non-default `explode: false` option. Useless thing imho 😅 but I'll look into this (we do have it the code).

dmitrysteblyuk: > is this a must have for you?

no 😄 i didn't bring it up. Only agreed it's a good thing to address since we don't have a standard implementation for this in our services. Splitting by `,` goes only so far.

packages/oats-runtime/src/make.ts
sakari: mmm. people have reservations on doing major version updates. This might conceivably break cases where there is an oneOf  `array | string` as the type to work around this limitation of array handling (or number handling for that matter).

dmitrysteblyuk: i'll check how `array | string` will work. I'd assume, first come - first served. So if array is first (`array | string`), it'll always convert to an array for a single element. If it's a string first (`string | array`), nothing should change.

packages/oats-runtime/src/make.ts
dmitrysteblyuk: i just checked, we're not supporting `explode=false` for queries. So we're only supporting the standard serialization.
There's this line
https://github.com/smartlyio/oats/blob/master/packages/oats/src/generate-types.ts#L244
but i'm not sure what it does because `explode` default value is true, so what is this check for? 🤔

packages/oats-runtime/src/make.ts
sakari: Would we be ok by making this opt in as in https://github.com/smartlyio/oats/pull/246

packages/oats-axios-adapter/src/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+    return '';
+  }
+  if (params instanceof URLSearchParams) {
+    return params.toString();
+  }
+  const queryParams = new URLSearchParams();
+  Object.keys(params).forEach(key => {
+    const value = (params as Record<string, unknown>)[key];
+    for (const item of ([] as unknown[]).concat(value)) {
+      queryParams.append(key, String(item));
```
taha-k, line 20: Hey can you explain how this works? Like if the object is {key1: [v1, v2, v3]}, then it seems this would resolve as `key1=v1&key1=v2&key1=v3`

packages/oats-runtime/src/make.ts
sakari: I would imagine that it deals with exploding the query params to an object. So you would have a single query param of type object and the incoming query object from the adapter is matched to that.

packages/oats-runtime/src/make.ts
```diff
@@ -198,10 +211,31 @@ export function makeString(
...
+  isInteger: boolean,
+  min?: number,
+  max?: number
+): Maker<unknown, number> {
+  return (x, { parseNumericStrings = false } = {}) => {
+    let value: number;
+    if (typeof x === 'number') {
+      value = x;
+    } else if (typeof x === 'string' && parseNumericStrings) {
+      value = parseFloat(x);
```
sakari, line 230: are we concerned that `parseFloat('1a') === 1`

packages/oats-runtime/src/make.ts
```diff
@@ -198,10 +211,31 @@ export function makeString(
...
 }
 
-export function makeNumber(min?: number, max?: number): Maker<number, number> {
-  return (value: any) => {
-    if (typeof value !== 'number') {
-      return getErrorWithValueMsg('expected a number', value);
+export function makeNumber(min?: number, max?: number) {
+  return makeFloatOrInteger(false, min, max);
+}
+export function makeInteger(min?: number, max?: number) {
```
sakari, line 217: separating integers from numbers seems to be the only change now that might break things. Although it is more of a bug fix.

packages/oats-runtime/src/make.ts
dmitrysteblyuk: > So you would have a single query param of type object
on server side query param is either a string or an array of strings, so how can it support a query param of type object?

packages/oats-runtime/src/make.ts
dmitrysteblyuk: perhaps `parseFloat(x) === Number(x)` should be a good enough way to make sure we have valid numbers in decimal format.

sakari: wouldn't using `value = Number(x)` do the thing without having to deal with parseFloat?

dmitrysteblyuk:
https://github.com/smartlyio/oats/pull/246#issuecomment-935595354
should we maybe deprecate the original generated `router`?

Screenshot 2021-10-07 at 08 16 11

and have these features (parse/array-convert support) in a new `createRouter()` without opt-out option?
Bc the only thing they break is `array | string` schema for a query param and there's no valid use case for it afaics. `array` type for multiple values, `string` for a single value should always work better.

packages/oats-runtime/src/make.ts
dmitrysteblyuk: > and breaking places where accidentally integer is used for type where number should be used (feels like a bug fix).

yes, good point - it could've been used accidentally (and worked well until this release), so it'll be a breaking change after all

dmitrysteblyuk:
Example of the new generated server router (with new features enabled by default).

Screenshot 2021-10-07 at 18 06 54

dmitrysteblyuk:
builds failing due to npm outag
Screenshot 2021-10-07 at 19 25 43
e

sakari (approved):

sakari (changes requested):
meh. a test breaks in my very important project. I have to figure out if that is a regression on this pr or something not being done correctly in there. Anyway that will take until next week when I have time. 

Consider this approved and I'll take this to the finishing line next week.

PR #246, opened 2021-10-05
allow parameters to server handler making

sakari:


dmitrysteblyuk:
does it cover client side as well? if these options enabled only for the server, won't request validation on client side fail?

dmitrysteblyuk:
also i think it is missing additional option for the `driver.generate` (I can add myself if you like). No many use `createHandlerFactory()` manually i think.

dmitrysteblyuk:
also should we deprecate call signature where these options (for query/path params) are not passed? 🤔

sakari:
on client side the types already force us to pass the correctly typed value like a number when expected. I'm not sure why we would want to allow passing a string there and then converting it to a number -- if I understood your question right. 

Doing something to encode the parameters correctly though might be nice instead of assuming that the adapter does something useful with it. With numbers and strings I think there might be a good chance it might do something useful. Not sure about arrays.

sakari:
> option for the driver.generate

I'would maybe just add something like to the generated source code and not have it as a generator option

```
export function routerWithOptions(opts: oar.server.HandlerOptions): oar.server.HandlerFactory<Endpoints> {
  return oar.server.createHandlerFactory(endpointHandlers, opts);
}
```

dmitrysteblyuk:
> on client side the types already force us to pass the correctly typed value like a number when expected

my bad, it is covered nicely then

dmitrysteblyuk:
> I'would maybe just add something like to the generated source code and not have it as a generator option

i'll take care of this

dmitrysteblyuk:
i'd only change then names
```
export type HandlerOptions = {
  queryMakeOptions?: MakeOptions;
  paramsMakeOptions?: MakeOptions;
};
```

PR #250, opened 2021-10-14
Feat/enable generated type prefix and suffix

innolab-xsd:
**WHY**
Structural types are prefixed with `ShapeOf` in Oats generation and nominal class are neither prefixed nor suffixed.  

**WHAT**
- Add a prop in driver and options: `enableTypeManipulation`
- When generating server and types, not prefix structural type with `ShapeOf` and suffix nominal type with `WithBrand` if `enableTypeManipulation=true`.
- Add simple test for type manipulation.

### DIFF
--- a/packages/oats/src/generate-server.ts
+++ b/packages/oats/src/generate-server.ts
@@ -202,7 +202,10 @@ function generateClientTree(
             undefined,
             tree.param.name,
             undefined,
-            ts.createKeywordTypeNode(ts.SyntaxKind.StringKeyword)
+            ts.createUnionTypeNode([
+              ts.createKeywordTypeNode(ts.SyntaxKind.StringKeyword),
+              ts.createKeywordTypeNode(ts.SyntaxKind.NumberKeyword)
+            ])
           )
         ],
         generateClientSpecType(opts, tree.param.tree)
@@ -373,6 +376,17 @@ function generateHandler(opts: Options) {
   );
 }
 
+function generateRouterJSDoc() {
+  return ts.createJSDocComment(undefined, [
+    ts.createJSDocTag(
+      ts.createIdentifier('deprecated'),
+      'Use `createRouter()` instead. ' +
+        'It supports "number", "integer", "boolean" and "array" types in query parameters ' +
+        'and numeric types in path parameters.'
+    )
+  ]);
+}
+
 export function generateRouter() {
   return ts.createVariableStatement(
     [ts.createModifier(ts.SyntaxKind.ExportKeyword)],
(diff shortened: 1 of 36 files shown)


packages/oats/src/generate-types.ts
```diff
@@ -227,24 +228,25 @@ export function run(options: Options) {
 
   function generateQueryType(
     op: string,
+    opts: Options,
```
sakari, line 231: do you need to pass `opts` around here? The functions are already in a closure that has it

packages/oats/src/driver.ts
```diff
@@ -42,6 +42,7 @@ export interface Driver {
     security?: UnsupportedFeatureBehaviour;
   };
   forceGenerateTypes?: boolean; // output the type file even if it would have been already generated
+  enableTypeManipulation?: boolean; // manipulate generation of structural and nominal types
```
sakari, line 45: could we have this as `typename?: (kind: shape | value, name: string) => string` as we are not actually doing anything else than renaming the types. We could also maybe wrap oautil.typenamify to do the type mapping as needed to make the code a bit less invasive.

another question is how this should work when importing types from other files? Having this setting in one file requires similar setting when generating the other files.

sakari: one way to do this would be to hook this up with the name resolvers if needed. If there is need for more finessing we could I guess require writing their own resolver to deal with the name mapping. However in any case I don't think we should not after resolution do any mapping with the name

packages/oats/src/driver.ts
```diff
@@ -106,7 +107,7 @@ export function generateFile(opts?: GenerateFileOptions): types.Resolve {
     return {
       importAs: moduleName,
       importFrom: generatedFile,
-      name: refToTypeName(localName),
+      name: options.nameMapper?.(refToTypeName(localName), 'value') ?? refToTypeName(localName),
```
sakari, line 110: how do you know that the context needing this is a 'value' and not a 'shape'?

Xiaolei-Shawn: To my understanding, while resolving reference, the value class is preferably used. Given the mapping function, it would make sense to mapping it as `value`, IMHO. We don't probably use either `shape` or `reflection` here, do we?

sakari: I think that we do need to resolve references to shapes also. and to reflection types for that matter. Both can be composed of references.

Xiaolei-Shawn: How do use `shape` or `reflection` as ref 🤔 ?
Well, if that is the case, we could pass the kind in `GenerateFileOptions` so that resolver could know in which kind the name should be resolved. Does it make sense to you?

packages/oats/src/util.ts
```diff
@@ -17,11 +17,23 @@ export function nonNullableClass(name: string) {
...
     name = 'Type' + name;
   }
-  return capitalize(name);
+  const capitalizedName = capitalize(name);
+  switch (nameKind) {
+    case 'shape':
+      return nameMapper?.(capitalizedName, nameKind) ?? 'ShapeOf' + capitalize(name);
+    case 'value':
+      return nameMapper?.(capitalizedName, nameKind) ?? capitalizedName;
+    case 'reflection':
```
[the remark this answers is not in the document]
Xiaolei-Shawn: Introduce local name mapper which is taken into use when custom name mapper is not present in drive option.

packages/oats/src/generate-types.ts
```diff
@@ -1497,7 +1497,7 @@ export function run(options: Options) {
   }
 
   function resolveRefToTypeName(ref: string): { qualified?: ts.Identifier; member: string } {
-    const external = options.resolve(ref, options);
+    const external = options.resolve(ref, options, 'value'); //TODO: value of 'kind' in the context is unsure
```
Xiaolei-Shawn, line 1500: How do we determine kind of ref in this context, @sakari ?

sakari: I'd say that you have to pass it from the caller of `resolveRefToTypeName`

packages/oats/src/driver.ts
```diff
@@ -56,17 +63,22 @@ function defaultResolve() {
   return undefined;
 }
 
-export function localResolve(ref: string) {
+export function localResolve(ref: string, options: types.Options, kind?: NameKind) {
   if (ref[0] === '#') {
-    return { name: refToTypeName(ref) };
+    return {
+      name:
+        kind && options.nameMapper
```
sakari, line 70: now this does not add the default prefixes if namemapper or kind is not given. Isn't that a mistake

test/name-mapping/custom/driver.ts
```diff
@@ -0,0 +1,30 @@
...
+import * as process from 'process';
+
+process.chdir(__dirname);
+const nameMapper = (name: string, kind: util.NameKind) => {
+  if (kind === 'value') {
+    return 'Value' + name;
+  } else if (kind === 'reflection') {
+    return 'Reflection' + name;
+  } else {
+    return name;
```
sakari, line 11: make make sense to add a prefix here also to ensure that shapes get processed

test/name-mapping/custom/driver.ts
```diff
@@ -0,0 +1,30 @@
+import { driver, util } from '@smartlyio/oats';
+import * as process from 'process';
+
+process.chdir(__dirname);
+const nameMapper = (name: string, kind: util.NameKind) => {
+  if (kind === 'value') {
```
sakari, line 6: even though there is a type it would be nice to have `else if kind === 'shape' ... else assert(false)` or something like that to ensure that we get the correct kind values.

packages/oats/src/generate-server.ts
```diff
@@ -53,36 +53,46 @@ function generateMethod<S extends oas.OperationObject>(
 
   const headers = ts.createTypeReferenceNode(
     fromTypes(
-      (opts.shapesAsRequests ? 'ShapeOf' : '') +
-        oautil.typenamify(oautil.endpointTypeName(schema, path, method, 'headers'))
+      opts.nameMapper(
```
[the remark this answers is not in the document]
Xiaolei-Shawn: `utils.typenamify()` does capitalization only but nothing else. I think capitalisation is done in default mapper so default behaviour should not change, right?

packages/oats/src/driver.ts
Xiaolei-Shawn: Kind of. If we don't do any manual `type` concatenation to resolved type name, a default prefix `type` is needed here.

packages/oats/src/generate-server.ts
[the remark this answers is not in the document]
sakari: oh. well then its ok. do we use typenamify at all?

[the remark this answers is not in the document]
Xiaolei-Shawn: Yes, for instance, `refToTypeName` is calling it.

PR #273, opened 2021-11-16
Prevent status code override in Koa adapter

astepantsov:


sakari (approved):
does this work?

astepantsov:
The test for this use case is now in place. Hopefully I've put it in the right place. 😅

sakari:
good enough

PR #280, opened 2021-12-08
Support cyclical dependencies

### LINKED ISSUE #140: oats-fast-check error about "cannot read property 'definition' of undefined" when there are circular references between spec files
when using oatsFastCheck to generate a value from a type there might be an error about not being able to get 'description' from an undefined value.

what do?
 - fail when trying to follow circular dependencies when emitting the code
 - figure out why this is an issue and fix

sakari:
Get rid of Object.assign and use functions for references instead.

should fix #140 

released as `@4.0.0-alpha.2` for testing

### DIFF
--- a/packages/oats-fast-check/src/generator.ts
+++ b/packages/oats-fast-check/src/generator.ts
@@ -67,7 +67,7 @@ export class GenType extends fc.Arbitrary<any> {
       assert.fail('todo intersection value generation');
     }
     if (type.type === 'named') {
-      return named(type.reference);
+      return named(type.reference());
     }
     if (type.type === 'null') {
       return fc.constant(null);
--- a/packages/oats/src/generate-types.ts
+++ b/packages/oats/src/generate-types.ts
@@ -833,18 +833,17 @@ export function run(options: Options) {
       return ts.createObjectLiteral(
         [
           ts.createPropertyAssignment('type', ts.createStringLiteral('named')),
-          ts.createPropertyAssignment('reference',
+          ts.createPropertyAssignment(
+            'reference',
             ts.factory.createArrowFunction(
               undefined,
               undefined,
               [],
               undefined,
               ts.factory.createToken(ts.SyntaxKind.EqualsGreaterThanToken),
-              ts.factory.createBlock(
-                [ts.factory.createReturnStatement(type)],
-                false
-              )
-            ))
+              ts.factory.createBlock([ts.factory.createReturnStatement(type)], false)
+            )
+          )
         ],
         true
       );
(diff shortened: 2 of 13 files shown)


packages/oats-fast-check/src/generator.ts
```diff
@@ -67,7 +67,7 @@ export class GenType extends fc.Arbitrary<any> {
       assert.fail('todo intersection value generation');
     }
     if (type.type === 'named') {
-      return named(type.reference);
+      return named(type.reference());
```
sakari, line 70: we resolve the reference to a reflection type in runtime instead of at startup to avoid issues with cyclical dependencies breaking things

packages/oats/src/generate-types.ts
```diff
@@ -417,9 +417,6 @@ export function run(options: Options) {
         const endpoint: oas.OperationObject = schema.paths[path][method];
         oautil.errorTag(`in ${method.toUpperCase()} ${path} query`, () =>
           response.push(
-            generateNamedTypeDefinitionDeclaration(
```
sakari, line 420: do not need to generate separate assign and declaration for reflection types. Instead we generate a declaration and assign it in the same expression

packages/oats/src/generate-types.ts
```diff
@@ -575,10 +560,24 @@ export function run(options: Options) {
       ],
       ts.createIdentifier('reflection'),
       undefined,
-      ts.createTypeReferenceNode(fromLib('reflection', 'NamedTypeDefinition'), [
+      ts.createTypeReferenceNode(fromLib('reflection', 'NamedTypeDefinitionDeferred'), [
```
sakari, line 563: this is `() => NamedTypeDefinition`

test/cyclical-dependencies/cyclical.spec.ts
```diff
@@ -0,0 +1,19 @@
+import * as first from './tmp/first.types.generated';
+
+describe('cyclical test', () => {
+  it('allows cyclically depending objects', () => {
+    expect(
+      first.typeFirstObject
```
sakari, line 6: seems to work

packages/oats/src/generate-types.ts
ganymedes01: What if we generated 
```ts
public static reflection = () => { return typeNonNullableNullable; };`
```

instead of:
```ts
public static reflection: oar.reflection.NamedTypeDefinitionDeferred<NonNullableNullable> = () => { return typeNonNullableNullable; };
```

The type definition there seems redundant since `typeNonNullableNullable` will always be typed correctly. Normally I wouldn't care, but if we're trying to reduce file size 🤷

ganymedes01 (approved):
LGTM!
tested and it seems to be working fine 👍

packages/oats/src/generate-types.ts
sakari: the type should get erased when building to js so it should not matter. OTOH doesn't ts require exported functions to be explicitly typed?

PR #286, opened 2021-12-15
fix recursion limit error with readonlyarrary shapes

sakari:


packages/oats-runtime/src/runtime.ts
```diff
@@ -24,6 +24,8 @@ export type ShapeOf<A> = A extends Scalar
   ? unknown
   : A extends Array<infer Item>
   ? Array<ShapeOf<Item>>
```
ganymedes01, line 26: We can drop this as `A extends ReadonlyArray<infer Item>` will catch it
```suggestion
```

ganymedes01 (approved):

rikutiira (approved):

PR #287, opened 2021-12-16
Fix discriminators with falsy enum values

ganymedes01:


sakari (approved):

sakari (changes requested):
the test says that this does not fix things

```
- option 1
            : No value matching discriminator key enabled
```

probably because you do not pass `enabled` in the test

packages/oats-runtime/test/make-from-reflection.spec.ts
```diff
@@ -151,6 +151,58 @@ describe('named', () => {
...
+                enum: [true]
+              },
+              required: true
+            }
+          }
+        }
+      ]
+    };
+    const fun = make.fromReflection({ type: 'union', options: [atype] });
+    expect(fun({ tag: true, name: 'aaa' }).success()).toEqual({ tag: false, name: 'aaa' });
```
sakari, line 200: 👀 are you sure about the `tag` instead of `enabled` here?

ganymedes01: copy pasting  is hard

sakari (approved):

PR #291, opened 2021-12-29
For Oats packages used in the browser: Make dependency on assert explicit

davissorenson:
Webpack v5 and later do not polyfill Node packages like `assert`. It is possible to manually add resolutions for those packages in the Webpack config, but it's much easier for end users if we just include those packages in our dependencies.

Packages which are used locally, like `oats`, do not need this dependency, as they are run in Node which includes `assert`.

### DIFF
--- a/packages/oats-axios-adapter/package.json
+++ b/packages/oats-axios-adapter/package.json
@@ -22,6 +22,7 @@
     "typescript": "^4.2.0"
   },
   "dependencies": {
+    "assert": "^2.0.0",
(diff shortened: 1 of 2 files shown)


sakari:
I wonder if it would be better to avoid using assert in these packages? what does including `assert` cost us?

sakari:
good catch -- however I think we should prevent import of any node modules, not just deal with assert. So I made https://github.com/smartlyio/oats/pull/292 that should deal with this also

PR #309, opened 2022-03-11
add option to omit unknown index type for additionalProperties

sakari:
`additionalproperties: true` results in `[key:string]: unknown` signatures in objects. This is tedious if the additional prop is there to allow non-breaking migrations. Lets allow omitting the unknow signatures.

### DIFF
--- a/packages/oats/src/driver.ts
+++ b/packages/oats/src/driver.ts
@@ -1,17 +1,17 @@
 import * as fs from 'fs';
 import * as yaml from 'js-yaml';
 import * as types from './generate-types';
+import { AdditionalPropertiesIndexSignature, Resolve } from './generate-types';
 import * as server from './generate-server';
 import * as path from 'path';
 import * as oas from 'openapi3-ts';
 import {
-  UnsupportedFeatureBehaviour,
-  refToTypeName,
   capitalize,
+  NameKind,
   NameMapper,
-  NameKind
+  refToTypeName,
+  UnsupportedFeatureBehaviour
 } from './util';
-import { Resolve } from './generate-types';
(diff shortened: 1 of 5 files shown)


packages/oats/src/driver.ts
```diff
@@ -59,6 +59,12 @@ export interface Driver {
    *  index signature accesses to have implicit undefined type so we can let the caller decide on the level of safety they want.
    * */
   emitUndefinedForIndexTypes?: boolean;
+  /** If 'AdditionalPropertiesIndexSignature.emit' or not set emit
+   * `[key: string]: unknown`
+   * for objects with `additionalProperties: true` or no additionalProperties set
+   * NOTE: the resulting runtime value may still have extra properties even if those are not visible in the type
+   * */
+  unknownAdditionalPropertiesIndexSignature?: AdditionalPropertiesIndexSignature;
```
sakari, line 67: this here is the opt in

packages/oats/src/generate-types.ts
```diff
@@ -145,6 +149,9 @@ export function run(options: Options) {
       return;
     }
     if (additional === true || additional == null) {
+      if (options.unknownAdditionalPropertiesIndexSignature === AdditionalPropertiesIndexSignature.omit) {
```
sakari, line 152: if the flag is on do not emit the index signature for true or omitted additionalproperties

rikutiira (approved):
LGTM 👍

sakari:
canary at @smartlyio/oats@4.1.1-alpha.14

ganymedes01 (approved):

PR #315, opened 2022-03-24
AETHER-2101 implement runtime redirect

dmitrysteblyuk:
This implements a "runtime.redirect()" helper. It returns the same "server.Response" object, but checks for redirect status and sets Location header.

### DIFF
--- a/packages/oats-koa-adapter/index.ts
+++ b/packages/oats-koa-adapter/index.ts
@@ -59,7 +59,10 @@ function adapter<StateT, CustomT, RequestContext>(
         });
         ctx.body = result.value.value;
         ctx.status = result.status;
-        ctx.set('Content-Type', result.value.contentType);
+
+        if (result.value.contentType !== runtime.noContentContentType) {
+          ctx.set('content-type', result.value.contentType);
+        }
         ctx.set(result.headers);
       }
     );
--- a/packages/oats-koa-adapter/index.ts
+++ b/packages/oats-koa-adapter/index.ts
@@ -59,6 +59,7 @@ function adapter<StateT, CustomT, RequestContext>(
         });
         ctx.body = result.value.value;
         ctx.status = result.status;
+        ctx.set('Content-Type', result.value.contentType);
         ctx.set(result.headers);
       }
     );
--- a/packages/oats-runtime/package.json
+++ b/packages/oats-runtime/package.json
@@ -23,6 +23,10 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
+    "@types/encodeurl": "^1.0.0",
+    "@types/escape-html": "^1.0.1",
+    "encodeurl": "^1.0.2",
+    "escape-html": "^1.0.3",
     "lodash": "^4.17.20"
   },
   "keywords": [
--- a/packages/oats-runtime/package.json
+++ b/packages/oats-runtime/package.json
@@ -23,9 +23,7 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
-    "@types/encodeurl": "^1.0.0",
     "@types/escape-html": "^1.0.1",
-    "encodeurl": "^1.0.2",
     "escape-html": "^1.0.3",
     "lodash": "^4.17.20"
   },
--- a/packages/oats-runtime/package.json
+++ b/packages/oats-runtime/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "yarn clean:dist && rm -f ./tmp/*.ts",
     "clean:dist": "rm -rf ./dist",
-    "prebuild": "yarn clean",
+    "prebuild": "yarn clean && yarn tsc --project tsconfig.test.json --noEmit",
     "build": "yarn tsc",
     "lint": "eslint --max-warnings=0 --ext .ts src test"
   },
--- a/packages/oats-runtime/package.json
+++ b/packages/oats-runtime/package.json
@@ -23,8 +23,8 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
-    "@types/escape-html": "^1.0.1",
-    "escape-html": "^1.0.3",
+    "@types/escape-html": "1.0.1",
+    "escape-html": "1.0.3",
     "lodash": "^4.17.20"
   },
   "keywords": [
--- a/packages/oats-runtime/src/redirect.ts
+++ b/packages/oats-runtime/src/redirect.ts
@@ -1,0 +1,65 @@
+import type { Response } from './server';
+import encodeUrl = require('encodeurl');
+import escapeHtml = require('escape-html');
+
+const REDIRECT_STATUSES = [300, 301, 302, 303, 305, 307, 308] as const;
+const DEFAULT_REDIRECT_STATUS = 302;
+const TEXT_HTML_CONTENT_TYPE = 'text/html';
+const TEXT_PLAIN_CONTENT_TYPE = 'text/plain';
+
+export type RedirectStatus = typeof REDIRECT_STATUSES[number];
+
+export function redirect<
+  Status extends RedirectStatus = typeof DEFAULT_REDIRECT_STATUS,
+  ContentType extends string = typeof TEXT_HTML_CONTENT_TYPE,
+  Value = string
+>(
+  url: string,
+  options: {
+    status?: Status;
+    contentType?: ContentType;
+    value?: Value;
+  } = {}
+): Response<Status, ContentType, Value, Record<string, any>> {
+  const { status = DEFAULT_REDIRECT_STATUS as Status } = options;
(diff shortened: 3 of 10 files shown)


dmitrysteblyuk:
Also I was thinking we could make this more secure by implementing `runtime.setAllowedRedirectUrls()`. The only concern is that `runtime.setAllowedRedirectUrls` can be called already in some node module.

In this case smth like this could work `runtime.setAllowedRedirectUrls((currentUrls) => [...currentUrls, 'https://example.com/also-allowed'])`

packages/oats-runtime/src/redirect.ts
```diff
@@ -0,0 +1,72 @@
...
+  ContentType extends string = typeof TEXT_HTML_CONTENT_TYPE,
+  Value = DefaultValue<ContentType>
+>(
+  url: string,
+  options: {
+    status?: Status;
+    contentType?: ContentType;
+    value?: Value;
+  } = {}
+): Response<Status, ContentType, Value, Record<string, any>> {
```
sakari, line 31: why headers is `Record<string, any>` instead of `{ location: string }`?

packages/oats-runtime/src/redirect.ts
```diff
@@ -0,0 +1,72 @@
...
+
+type DefaultValue<ContentType extends string> = ContentType extends
+  | typeof TEXT_HTML_CONTENT_TYPE
+  | typeof TEXT_PLAIN_CONTENT_TYPE
+  | `${typeof TEXT_HTML_CONTENT_TYPE};${string}`
+  | `${typeof TEXT_PLAIN_CONTENT_TYPE};${string}`
+  ? string
+  : null;
+
+export function redirect<
```
sakari, line 20: I'm kinda wondering why we need all of this code when plain helpers around runtime.json and runtime.setHeaders would do for me. Kinda hesitant forcing the users hand so much.

What others think?

packages/oats-runtime/package.json
```diff
@@ -23,6 +23,10 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
+    "@types/encodeurl": "^1.0.0",
+    "@types/escape-html": "^1.0.1",
+    "encodeurl": "^1.0.2",
```
sakari, line 28: could we do with just https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI instead of adding a dependency?

packages/oats-runtime/src/redirect.ts
```diff
@@ -0,0 +1,72 @@
...
+  const {
+    contentType = TEXT_HTML_CONTENT_TYPE as ContentType,
+    value = getDefaultRedirectValue({ encodedUrl, contentType }) as unknown as Value
+  } = options;
+
+  return {
+    status,
+    value: { contentType, value },
+    headers: {
+      Location: encodedUrl
```
sakari, line 47: does this need to be encoded? to me it sounds like a task for the adapter if that is needed

packages/oats-runtime/package.json
dmitrysteblyuk: oh I didn't know about it, but i'll fix this

packages/oats-runtime/src/redirect.ts
dmitrysteblyuk: the way I see we should do it anyways so why do it in every adapter? besides in adapter it'll quite messy, as we need to modify Location header (and it's not even case sensitive).

packages/oats-runtime/src/redirect.ts
dmitrysteblyuk: right, I thought it didn't matter. Do we validate response headers so that they match openapi spec?

packages/oats-runtime/src/redirect.ts
dmitrysteblyuk: I thought it handy. It encodes url, sets Location header, sets response text or html (not sure if they're that needed tho, but koa and express for example have them). Also we can add functionality to whitelist only allowed redirect urls. Otherwise, it can be a security issue.

packages/oats-runtime/src/redirect.ts
sakari: my hunch is that eg koa does the right thing already for headers

packages/oats-runtime/src/redirect.ts
sakari: I think we have a type for the headers and we do validate

packages/oats-runtime/src/redirect.ts
dmitrysteblyuk: well koa does encode url in their redirect - https://github.com/koajs/koa/blob/master/lib/response.js#L270
so does express - https://github.com/expressjs/express/blob/947b6b7d57939d1a3b33ce008765f9aba3eb6f70/lib/response.js#L893

sakari: if you set a location or any other header for koa does it need encoding?

dmitrysteblyuk: > any other header for koa does it need encoding

idk maybe some encoding happens for all headers, but is it URI encoding? i can of course try to break stuff by setting non-encoded url and see what happens

dmitrysteblyuk: okay, I broke it by having Cyrillic in the url 😅 works with encodeURI tho

packages/oats-runtime/src/redirect.ts
dmitrysteblyuk: yeah probably not the best place for whitelisting to be. The better place will be openapi spec in case redirect uri is give in request parameters. In case it's hardcoded/coming from config it's already should be considered allowed.

sakari:
could you set up a test that tests that the location url gets correctly passed to the client from a server? You can add such a test to `test/koa-adapter/` which already spins up a server for this.

packages/oats-runtime/package.json
```diff
@@ -23,6 +23,8 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
+    "@types/escape-html": "^1.0.1",
+    "escape-html": "^1.0.3",
```
sakari, line 27: this has been last updated 2019. I feel that we could maybe do without the dependency

dmitrysteblyuk: i also thought so, but couldn't find anything. Last publish was 7 years ago right, but maybe it's just perfect? 😅
i mean it doesn't do anything but escapes quotes and ampersands

dmitrysteblyuk: we can do it ourselves of course:
```
function escape(htmlStr) {
   return htmlStr.replace('&', "&amp;")
         .replaceAll('<', "&lt;")
         .replaceAll('>', "&gt;")
         .replaceAll('"', "&quot;")
         .replaceAll("'", "&#39;");        

}
```

sakari: you wanna deal with left-pad all again 😬

dmitrysteblyuk: sure, i'll remove the anchor
i can even add `save-exact=true` to .npmrc

packages/oats-koa-adapter/index.ts
```diff
@@ -59,6 +59,7 @@ function adapter<StateT, CustomT, RequestContext>(
         });
         ctx.body = result.value.value;
         ctx.status = result.status;
+        ctx.set('Content-Type', result.value.contentType);
```
dmitrysteblyuk, line 62: this is unrelated to this PR fix, but otherwise tests were not passing

packages/oats-koa-adapter/index.ts
```diff
@@ -59,6 +59,10 @@ function adapter<StateT, CustomT, RequestContext>(
         });
         ctx.body = result.value.value;
         ctx.status = result.status;
+
+        if (result.value.contentType !== runtime.noContentContentType) {
+          ctx.set('content-type', result.value.contentType);
+        }
```
dmitrysteblyuk, line 65: this fix is unrelated to this PR, but without it tests are not passing

packages/oats-runtime/package.json
```diff
@@ -23,6 +23,8 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
+    "@types/escape-html": "1.0.1",
+    "escape-html": "1.0.3",
```
sakari, line 27: mmm. maybe this is ok as [koa depends](https://www.npmjs.com/browse/depended/escape-html) on it anyway 😁

dmitrysteblyuk: if it's a fixed version, we're quite safe as it's not possible to unpublish a package in npm anymore (unless it has no dependencies and there're only few downloads a week). So no left-pad scenario for us.

sakari (approved):

dmitrysteblyuk:
I wanted to write change logs still 😬

PR #324, opened 2022-04-12
fix: use encodeurl instead of encodeURI

dmitrysteblyuk:
Turned out `encodeURI` does double encoding of already encoded params. Both koa and express are using `encodeurl` package in their redirect implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 Prior to that, the content type was determined automatically by Koa.
 
+# 4.3.2
+
+## Bug Fixes
+
+### `encodeURI` is replaced with `encodeurl` for redirect url encoding.
+
+`encodeURI` does double encoding of already encoded query parameters. `runtime.redirect()` will encode url with `encodeurl` now. See https://github.com/pillarjs/encodeurl.
+
 # 4.1.0
 
 Provide option `emitUndefinedForIndexTypes` (default true for backwards compatibility) which can be set to 
--- a/packages/oats-runtime/package.json
+++ b/packages/oats-runtime/package.json
@@ -23,7 +23,9 @@
   },
   "dependencies": {
     "@smartlyio/safe-navigation": "^5.0.1",
+    "@types/encodeurl": "1.0.0",
     "@types/escape-html": "1.0.1",
+    "encodeurl": "1.0.2",
     "escape-html": "1.0.3",
     "lodash": "^4.17.20"
   },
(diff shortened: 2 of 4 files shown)


sakari:
> This function will not encode the "%" character unless it is not part of a valid sequence (%20 will be left as-is

This seems a bit magic to me. I would think it way simpler to say that the uri should be passed unencoded and leave things as is.

dmitrysteblyuk:
> Why do we need to pass in encoded uris?

If you construct a url with `new URL` or `new URLSearchParams` it'll be encoded and you cannot change this.

The problem is that url can be passed both encoded or not, hence this magic is in place. Without it I cannot see how this can be done without requiring user to specify whether their url is encoded already explicitly (in an additional option/arg).
Also afaics `encodeurl` works exactly as `new URL` does. The only difference is that URL works for full urls only, whereas `encodeurl` works also for relative urls.

dmitrysteblyuk:
As an alternative (and safer) option we can require url to be of `URL` type and not a string. This way we don't need encoding, but we lose support of relative urls. We don't want relative urls anyways if they're provided by the api user (bc checking whether url is relative or absolute is tricky), but redirecting to a trusted relative url will not be possible either.

sakari:
not allowing relative urls sounds kinda limiting. 

Could we do this with taking in 'string | URL' where string is required to be unescaped and Url is `new URL(..)`. However are you sure that URL does escaping as we want? I mean I tested creating one with `http://aaaa?%foo` and its toString was not changed in any way

dmitrysteblyuk:
>I mean I tested creating one with http://aaaa?%foo and its toString was not changed in any way

It is true that new URL and encodeurl produce a different result in case of `http://aaaa?%foo`, but when decoded it'll still be the same `%foo`. So basically if you navigate to `http://aaaa?%foo` or to `http://aaaa?%25foo`, the server will have the same query param.

> Could we do this with taking in 'string | URL' where string is required to be unescaped and Url is new URL(..)

I'm afraid this will be confusing for the end user. I'd rather rely on `encodeurl` so that whoever using it doesn't need to think about the intricacies of the implementation. It's commonly used so it its behavior won't come as a surprise.

sakari:
mkay, could you jsdoc this behaviour very clearly and create a changelog. What kind of version bump you think this warrants?

dmitrysteblyuk:
> could you jsdoc this behaviour very clearly and create a changelog

roger that.

> What kind of version bump you think this warrants?

A patch as it fixes double encoding which wasn't the expected behavior.

sakari (approved):

PR #344, opened 2022-07-14
Pass parameterised path down to adapters

ganymedes01:
currently there's no way to access the parameterised path in the adapter level.
if one wishes to log outbound requests without path parameters one would need to write code to replace them. oats has the value ready, so let's just make it accessible

### DIFF
--- a/packages/oats-runtime/src/client.ts
+++ b/packages/oats-runtime/src/client.ts
@@ -28,7 +28,16 @@ export interface ClientSpec {
   readonly [part: string]: ClientSpec | PathParam | ClientEndpoint<any, any, any, any>;
 }
 
-export type ClientAdapter = server.SafeEndpoint;
+export type RequestContext = { parameterised_path: string };
+
+export type ClientAdapter = server.Endpoint<
+  Headers | undefined,
+  server.Params | undefined,
+  server.Query | undefined,
+  server.RequestBody<any> | undefined,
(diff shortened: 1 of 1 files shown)


packages/oats-runtime/src/client.ts
```diff
@@ -196,7 +205,7 @@ function makeMethod(adapter: ClientAdapter, handler: server.Handler, pathParams:
       headers: safe(ctx).headers.$,
       query: safe(ctx).query.$,
       body: safe(ctx).body.$,
-      requestContext: {}
+      requestContext: { parameterised_path: handler.path }
```
sakari, line 208: so this would be used by say axios adapter for logging? if so are we missing that bit in this pr.

```suggestion
      requestContext: { path: handler.path }
```

ganymedes01: >  I assume it would be something like /path/{id}

correct

ganymedes01: > so this would be used by say axios adapter for logging? if so are we missing that bit in this pr.

afaik we don't do any logging on metric collection in the axios/et al adapters, am I missing it?

sakari: we are not. I was more thinking about how would one use the request ctx there. that would have made the usage a bit clearer. But if you promise that this is the best we can do then whatevs

sakari (approved):

PR #350, opened 2022-08-02
propagate types through any even with surprising clone

sakari:
allOf with nested objects did not serialize properly due to additionalprops: true losing the type


sakari:
canary at @smartlyio/oats-runtime@5.1.4-alpha.0

wovalle (approved):
Don't have much context but it LGTM!

ganymedes01 (approved):
LGTM 👍

PR #356, opened 2022-09-20
Ts 4 8

sakari:
the code gen got broke so need to update. Because we update minimum ts version we need to bump major. Ts 4.8 and ts 4.6 are not compatible wrt the changes.

A QQ: should we provide facade for ts < 4.8 to not have to bump major? Don't know how involved it would be.


ilkkao (approved):
looks all good to me.

ilkkao:
typescript update to 4.8 in projects that use oats will most likely always require some type fixing so it won't happen immediately. On the other hand oats is not under heavy development. In the worst case some important fix could be backported to the 5.x branch?

PR #358, opened 2022-09-26
Dont use deprecated stuff

ganymedes01:


packages/oats/src/driver.ts
```diff
@@ -162,8 +162,6 @@ export function generateFile(opts?: GenerateFileOptions): types.Resolve {
           header: options.header,
           generatedValueClassFile: generatedFile + '.ts',
           resolve: options.resolve,
-          externalOpenApiSpecs: options.externalOpenApiSpecs,
```
sakari, line 165: why we lose these?

sakari: ah those are deprecated in some previous release already

sakari (approved):
seems legit. Could you test this with a canary and add a changelog if everything seems ok. I think we need a major as you lost the oats deprecated stuff.

sakari:
added also a note that we are downwards incombatible with ts < 4.8

PR #389, opened 2023-03-01
fix: skip props with undefined value

ledancs:


packages/oats-runtime/src/make.ts
```diff
@@ -457,6 +457,10 @@ export function makeObject<
       result[index] = propResult.success();
     }
     for (const index of Object.keys(value)) {
+      // do not consider props that have undefined value
```
sakari, line 460: do we want to prevent props with undefined values when additionalProperties: false?

ledancs: mmm does it matter if we have additionalProperties true or false? should this be done consistently regardless? 

from the developer point of view, I suppose if additionalProperties is true then we can leave them be, WDYT?

sakari: additionalProperties: false means that no additional properties are allowed. Might be that we do not consider undefined valued props as additional and are ok to ignore and drop those.

sakari: good point though: it theoretically might be a breaking (and possibly tricky to spot) change if additionalProperties: true is set and we would start omitting the props with undefined values. Somebody could be constructing values using `.make({ prop: undefined})` and dropping `prop` would change the resulting value. Not an issue when json serializing but an issue for `Object.keys(theMadeValue)`.

packages/oats-runtime/test/make.spec.ts
```diff
@@ -20,6 +21,15 @@ describe('createMakerWith', () => {
     expect(result).toBeInstanceOf(TestClass);
     expect(result.a).toEqual(['a']);
   });
+
+  it('when additional properties is false then remove undefined values', () => {
```
sakari, line 25: 1) we can test just the makeObject call, which is I think *way* simpler as we do not need to define a TestClass
2) I think we should remove undefined when additionalProperties in the openapi spec is some schema or 'false' -- but not otherwise.

ledancs: 1. ok thanks for the tip, I will update the test.
2. would this be a good place to filter out the undefined values? 🤔 
```
      const propResult: Make<any> = additionalProp(value[index], opts);
      if (propResult.isError()) {
        return propResult.errorPath(index);
      } else if (value[index] === undefined) {
        // when a schema is provided as additional props we should also skip the undefined values
        continue;
      }
```

ledancs: mmm won't let me get away with just makeObject, it does not accept undefined as value
image

sakari (approved):
Could you add a CHANGELOG.md entry for the next minor version

ledancs:
> 

yes sir!

PR #391, opened 2023-03-15
Make oats tag truly private

ganymedes01:
generated class after this change:
```ts
export class Error extends oar.valueClass.ValueClass {
    readonly message!: string;
    readonly messageIndex!: number;
    // @ts-ignore tsc does not like unused privates
    readonly #__oats_value_class_brand_tag!: string;
    readonly [instanceIndexSignatureKey: string]: unknown;
    public constructor(value: ShapeOfError, opts?: oar.make.MakeOptions | InternalUnsafeConstructorOption) { super(); oar.instanceAssign(this, value, opts, buildError); }
    public static reflection: oar.reflection.NamedTypeDefinitionDeferred<Error> = () => { return typeError; };
    static make(value: ShapeOfError, opts?: oar.make.MakeOptions): oar.make.Make<Error> { const make = buildError(value, opts); if (make.isError())
        return oar.make.Make.error(make.errors);
    else
        return oar.make.Make.ok(new Error(make.success(), { unSafeSet: true })); }
}
```

### DIFF
--- a/packages/oats/src/generate-types.ts
+++ b/packages/oats/src/generate-types.ts
@@ -127,7 +127,7 @@ export function run(options: Options) {
   function generateOatsBrandProperty() {
     return ts.factory.createPropertyDeclaration(
       [ts.factory.createToken(ts.SyntaxKind.ReadonlyKeyword)],
-      ts.factory.createIdentifier(`#${oatsBrandFieldName}`),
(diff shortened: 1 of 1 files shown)


sakari (approved):
seems as legit as the rest of the code.

PR #392, opened 2023-03-15
actually drop undefined for unknown additionalprops

sakari:
in case additionalProperties: false we did not drop undefined props like we planned. Now we do.


ledancs:
mmm I missed the correct spot to do the check... good catch!
oh but why were the tests passing before?

ganymedes01 (approved):

ledancs (approved):

PR #394, opened 2023-03-17
avoid parsing objects if the object has already been parsed using the given type.

sakari:
this is slightly risky as we end up aliasing things that we previously did not.
 - the caller might get confused when objects start to be identical
 - oats might get confused as we enrich the values with type metadata (all parses of the value add to meta data for that single object even if those come from different contexts). Not sure if that is ok.


packages/oats/src/builder.ts
```diff
@@ -0,0 +1,21 @@
+import * as ts from 'typescript';
+import * as assert from 'assert';
+
+export function buildMethod(methodStr: string): ts.MethodDeclaration {
```
sakari, line 4: these allow writing ts code generators as strings

packages/oats/src/generate-types.ts
```diff
@@ -661,111 +662,21 @@ export function run(options: Options) {
...
-                )
-              ]
-            )
-          )
-        )
-      ])
-    );
+    const className = options.nameMapper(key, 'value');
+    const shapeName = options.nameMapper(key, 'shape');
+    return buildMethod(`
```
sakari, line 667: see. slightly less verbose.

packages/oats-runtime/src/make.ts
```diff
@@ -438,6 +438,12 @@ export function makeObject<
     if (typeof value !== 'object' || value == null || Array.isArray(value)) {
       return getErrorWithValueMsg('expected an object', value);
     }
+
+    // if the value been parsed already using the given object type
+    // return the value as is
+    if (type && getTypeSet(value)?.has(type)) {
+      return Make.ok(value);
+    }
     comparisorOrder ||= Object.keys(props);
```
ganymedes01, line 447: how will this work with `oneOf`s? specially in the case where a discriminator cannot be determined, wouldn't we still check each class until we find the one that the object was already an instance of?

sakari: we would check each branch of the oneOf -- however because of line 444 the checks should be fast. 

Or do I miss your point completely?

sakari: ah. now I got it. Yes we would still try to parse in each branch which is not great. We could add the oneOf type to the metadata for the parsed object to sidestep that I think 🤔 . Kinda would be *way* better to have discriminated objects as that would help in more cases.

packages/oats/src/generate-types.ts
ganymedes01: a better indented 👍

ganymedes01 (approved):
LGTM ✨

PR #402, opened 2023-03-29
If no headers are defined, generate type as `Record<string, string|undefined>`

ganymedes01:
If one header is defined the generated type is:
```ts
    readonly headers: {
        readonly etag: string;
        readonly [key: string]: string | undefined;
    };
```

but if no header is the defined the type is:
```ts
    readonly headers: {};
```

with this PR the generated type if no headers are defined will be:
```ts
    readonly headers: {
        readonly [key: string]: string | undefined;
    };
```
which, in my opinion, is the expected behaviour given the first scenario

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "client"
   ],
   "devDependencies": {
-    "@smartlyio/safe-navigation": "5.0.2",
     "@types/jest": "29.5.0",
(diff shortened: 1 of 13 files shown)


sakari (approved):
I disapprove the lack of panache and elegance in accessing properties

PR #406, opened 2023-04-05
Add `oats-fetch-adapter`

ganymedes01:
moved over from https://github.com/smartlyio/oats-fetch-adapter which was kinda abandoned so there was some heavy refactoring done as well


test/fetch-adapter/fetch-adapter.spec.ts
valentin-nemcev, line 1: I see you test that it sends stuff to server correctly, but I don't see tests that it receives stuff correctly, beyond 200 status?

ganymedes01: true. i’ll add those

packages/oats-fetch-adapter/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    const server = arg.servers[0];
+    const data = toRequestData(arg.body);
+    const url = new URL(server + arg.path);
+    const requestContentType = arg.body?.contentType;
+
+    Object.entries(arg.query ?? {})
+      .filter(([, value]) => !!value)
+      .forEach(([key, value]) => url.searchParams.append(key, `${value}`));
+
+    const response = await fetch(
```
valentin-nemcev, line 74: do you need to put credentials: "some-origin" or whatever you need to send cookies with fetch? And some other security stuff about cors? I always forget

valentin-nemcev: ok, nvm, it has defaults

sakari (approved):
I guess this is fine if it works

PR #416, opened 2023-04-28
Add body make options for client & server so clients can define how to handle values

smartly-petri:
Clients of `client` & `server` implementations can define what to do with unknown fields; to throw error or just drop fields.

### DIFF
--- a/packages/oats/src/generate-server.ts
+++ b/packages/oats/src/generate-server.ts
@@ -509,24 +509,6 @@ export function generateCreateRouter() {
                             ts.factory.createStringLiteral('fail')
                           ),
                           ts.factory.createBindingElement(
-                            undefined,
-                            'parseBooleanStrings',
-                            'bodyParseBooleanStrings',
-                            ts.factory.createTrue()
-                          ),
-                          ts.factory.createBindingElement(
-                            undefined,
-                            'parseNumericStrings',
-                            'bodyParseNumericStrings',
-                            ts.factory.createTrue()
-                          ),
-                          ts.factory.createBindingElement(
-                            undefined,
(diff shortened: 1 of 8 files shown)


packages/oats/src/generate-server.ts
```diff
@@ -498,6 +498,42 @@ export function generateCreateRouter() {
...
+                            'parseNumericStrings',
+                            'bodyParseNumericStrings',
+                            ts.factory.createTrue()
+                          ),
+                          ts.factory.createBindingElement(
+                            undefined,
+                            'allowConvertForArrayType',
+                            'bodyAllowConvertForArrayType',
+                            ts.factory.createTrue()
+                          ),
```
buttilda, line 528: i don't think we need to deconstruct these

smartly-petri: Yeah, no need at least for now. Removed 👍

packages/oats/src/generate-server.ts
```diff
@@ -565,7 +601,29 @@ export function generateCreateRouter() {
...
+                          ts.factory.createIdentifier('bodyParseBooleanStrings')
+                        ),
+                        ts.factory.createPropertyAssignment(
+                          'parseNumericStrings',
+                          ts.factory.createIdentifier('bodyParseNumericStrings')
+                        ),
+                        ts.factory.createPropertyAssignment(
+                          'allowConvertForArrayType',
+                          ts.factory.createIdentifier('bodyAllowConvertForArrayType')
+                        )
```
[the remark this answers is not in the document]
smartly-petri: Done

buttilda (approved):
LGTM

PR #418, opened 2023-05-03
Change drop option to behave the same way 'fail' would when matching to schema

smartly-petri:
This wouldn't be backwards compatible change as now we are loosening validation with "drop". 

Earlier we would have thrown an error (which some clients might be depending at the moment, hopefully not) if input matched multiple schemas with "drop" option. Now the behaviour would change to first check which object had the most matched properties and use it.

If multiple matched objects has the same amount of matched properties, we can't identify the best match and return the same error as earlier `multiple options match`

### DIFF
--- a/packages/oats-runtime/src/make.ts
+++ b/packages/oats-runtime/src/make.ts
@@ -352,7 +352,7 @@ export function makeOneOf(...options: any[]) {
             return error('multiple preferred options match');
           }
           preferredSuccess = mapped;
-        } else if (0 < success.length) {
+        } else if (success.length > 0) {
           if (opts?.unknownField === 'drop') {
             success.push(mapped);
           } else {
@@ -384,7 +384,7 @@ export function makeOneOf(...options: any[]) {
 }
 
 function isMultipleMatchesFound(mapped: any[]): boolean {
-  return 1 < mapped.length;
+  return mapped.length > 1;
 }
 
 function findBestMatch(mapped: any[]): any {
--- a/packages/oats-runtime/src/make.ts
+++ b/packages/oats-runtime/src/make.ts
@@ -342,7 +342,7 @@ export function makeOneOf(...options: any[]) {
     if (options.length === 0) {
       errors.push(error('no options given for oneof'));
     }
-    let success;
+    const success = [];
     let preferredSuccess;
     for (const option of options) {
       const mapped = option(value, opts);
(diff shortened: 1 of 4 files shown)


packages/oats-runtime/src/make.ts
```diff
@@ -352,24 +352,56 @@ export function makeOneOf(...options: any[]) {
             return error('multiple preferred options match');
           }
           preferredSuccess = mapped;
-        } else if (success) {
-          return error('multiple options match');
+        } else if (0 < success.length) {
```
taha-k, line 355: its really stupid nitpick but this reads easier in english
```suggestion
        } else if (success.length > 0) {
```

taha-k (approved):
Couldn't understand the core logic, but based on test cases looks good to me

packages/oats-runtime/src/make.ts
buttilda: yeah my brain froze for a second there when I saw it 😂 somehow feels like wrong syntax

smartly-petri: Fine, done 😂

packages/oats-runtime/src/make.ts
```diff
@@ -352,24 +352,56 @@ export function makeOneOf(...options: any[]) {
...
-      return preferredSuccess || success;
+    if (preferredSuccess) {
+      return preferredSuccess;
+    }
+
+    if (success.length === 1) {
+      return success[0];
     }
+
+    if (isMultipleMatchesFound(success)) {
```
sakari, line 376: could you motivate why we need this kind of shenanigans?

As is this feels very specific to what you happen to need which would lead me to think that maybe we could achieve the same using makeoptions or even some kind of middleware setup. I kinda would prefer having a time machine and not doing the whole drop thing altogether as it makes thinking about the behaviour way harder for me.

buttilda (approved):
LGTM

packages/oats-runtime/src/make.ts
smartly-petri: Sure thing.

We want to move to a state between s2s communications that all schemas between services are not always 1-to-1. This allows us having looser coupling between services; service A can start to send new property to service B that doesn't still have that definition in its own schema.

As we are using oats for parsing our requests & responses, then using oats types within our service boundaries, this way we can achieve the looser coupling with the guarantee that within our service boundaries objects only have properties defined in the schema itself.

This change would make `drop` behave the same way as `fail` does when matching to schemas which allows services to change the default behaviour of `fail` to `drop` without having to, possibly, change how their schemas are defined.

sakari: would something like https://github.com/smartlyio/oats/pull/419 do?

PR #421, opened 2023-05-09
Add possibility drop additional properties when client is handling response from server

smartly-petri:
Currently client can't drop additional properties even though client is created with:
```ts
client.createClient({
      validationOptions: { body: { unknownField: 'drop' } }
    })
```

Lets fix that, and let client also drop additional properties from server responses.

This currently works for:
* client sending request
* server receiving request

but was missing from client receiving response from the server.

### DIFF
--- a/packages/oats-runtime/src/client.ts
+++ b/packages/oats-runtime/src/client.ts
@@ -155,9 +155,10 @@ export interface OpTree<T> {
 function makeParam(
   adapter: ClientAdapter,
   param: { name: string; tree: OpTree<server.Handler> },
-  pathParams: string[]
+  pathParams: string[],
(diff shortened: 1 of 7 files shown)


packages/oats-runtime/src/client.ts
```diff
@@ -214,9 +215,9 @@ function fromTree(
   adapter: ClientAdapter,
   tree: OpTree<server.Handler>,
   pathParams: string[],
-  opts?: server.HandlerOptions
+  opts: server.HandlerOptions | undefined
```
buttilda, line 218: why this change?

buttilda (approved):

PR #441, opened 2023-07-20
fix empty headers object validation in case all declared headers are optional

dmitrysteblyuk:
Note, `ctx.headers` will be `{}` in case there are only optional headers declared but not sent. Before it was converted to `null` in this case.

### DIFF
--- a/packages/oats-runtime/src/server.ts
+++ b/packages/oats-runtime/src/server.ts
@@ -131,11 +131,8 @@ function voidify(value: object | undefined | null) {
 function cleanHeaders<H>(mode: Mode, maker: Maker<any, H>, headers: object) {
   // note: we now expect that on client side headers are type conforming so do not need to lowercase those
   // this is slightly breaking change is somebody has subverted the type checker
-  const normalized = voidify(mode === 'server' ? lowercaseObject(headers) : headers);
-  const acceptsNull = maker(null);
-  if (acceptsNull.isSuccess()) {
-    return acceptsNull.success();
-  }
+  const normalized = mode === 'server' ? lowercaseObject(headers) : headers;
+
   return maker(normalized, {
     unknownField: 'drop',
     convertFromNetwork: mode === 'server'
@@ -189,7 +186,7 @@ export function safe<
       method: ctx.method,
       servers: ctx.servers,
       op: ctx.op,
-      headers: serializeWhenClient(mode, cleanHeaders(mode, headers, ctx.headers)),
+      headers: serializeWhenClient(mode, cleanHeaders(mode, headers, ctx.headers ?? {})),
       // note: path params come to the adapter in network format always as those are gleaned from the path definition
       params: params(voidify(ctx.params), {
         ...validationOptions.params,
--- a/packages/oats-runtime/src/server.ts
+++ b/packages/oats-runtime/src/server.ts
@@ -129,6 +129,11 @@ function voidify(value: object | undefined | null) {
 }
 
 function cleanHeaders<H>(mode: Mode, maker: Maker<any, H>, headers: object) {
+  // If headers are not declared in the spec, return null.
+  const acceptsNull = maker(null);
+  if (acceptsNull.isSuccess()) {
+    return acceptsNull.success();
+  }
   // note: we now expect that on client side headers are type conforming so do not need to lowercase those
   // this is slightly breaking change is somebody has subverted the type checker
   const normalized = mode === 'server' ? lowercaseObject(headers) : headers;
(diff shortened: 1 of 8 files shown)


packages/oats-runtime/src/server.ts
```diff
@@ -131,11 +131,8 @@ function voidify(value: object | undefined | null) {
 function cleanHeaders<H>(mode: Mode, maker: Maker<any, H>, headers: object) {
   // note: we now expect that on client side headers are type conforming so do not need to lowercase those
   // this is slightly breaking change is somebody has subverted the type checker
-  const normalized = voidify(mode === 'server' ? lowercaseObject(headers) : headers);
-  const acceptsNull = maker(null);
-  if (acceptsNull.isSuccess()) {
-    return acceptsNull.success();
-  }
+  const normalized = mode === 'server' ? lowercaseObject(headers) : headers;
```
dmitrysteblyuk, line 134: I can still keep the old before (acceptsNull check which i removed). Just removing voidify() will fix the bug with the empty headers.
One thing i don't like about the current implementation is that headers value is typed as `void`, but its value is `null` (this is in case no header parameters is specified in the spec). I'm not a TS expert, but to my knowledge undefined can be used as void type, not null.

dmitrysteblyuk: Btw same is true for `ctx.params` - `null` value, `void` type.

packages/oats-runtime/src/server.ts
```diff
@@ -131,11 +131,8 @@ function voidify(value: object | undefined | null) {
 function cleanHeaders<H>(mode: Mode, maker: Maker<any, H>, headers: object) {
   // note: we now expect that on client side headers are type conforming so do not need to lowercase those
   // this is slightly breaking change is somebody has subverted the type checker
-  const normalized = voidify(mode === 'server' ? lowercaseObject(headers) : headers);
-  const acceptsNull = maker(null);
```
sakari, line 135: I don't understand this. We return null no matter what the headers are if null is accepted? That seems just wrong

sakari:
nice improvement. Can we test this in some larger project to check that there is not anything that we are missing?

packages/oats-runtime/src/server.ts
dmitrysteblyuk: > if null is accepted

null is only accepted if there are no headers. Bc otherwise, the schema will be of type `object`, not `void`. Basically we're relying on an implementation detail here (i.e. we don't have a schema that accepts both null and objects).

packages/oats/src/generate-types.ts
```diff
@@ -16,8 +16,11 @@ interface ImportDefinition {
 }
 
 // bit of a lie here really
-const voidSchema: oas.SchemaObject = { type: 'void' as any };
-const emptyObjectSchema: oas.SchemaObject = { type: 'object', additionalProperties: false };
+const VOID_SCHEMA: oas.SchemaObject = Object.freeze({ type: 'void' as any });
```
sakari, line 19: do we need freezing really? Asking as we are not using freeze anywhere else so its a bit of a surprise to see it here.

dmitrysteblyuk: I felt like for constants it's a nice way to ensure they will stay constant. If none of the functions below mutates these objects, the behavior will be the same (with freeze or without), otherwise, it's a big no no, so an error will be thrown.

sakari: does node (in our usage) really throw and not fail silently. Testing this in node repl does not seem to throw.

dmitrysteblyuk: alternatively, we can just replace a constant with factory:
```suggestion
const createVoidSchema = (): oas.SchemaObject => ({ type: 'void' as any });
```
?

sakari: isn't that ts strict not "strict"?

sakari: well a factory would do but would maybe be simpler to just not care

dmitrysteblyuk: > isn't that ts strict not "strict"?

yes, it's `alwaysStrict` that emits 'use strict' - https://www.typescriptlang.org/tsconfig#alwaysStrict

> but would maybe be simpler to just not care

mm, sure. I felt like it can be a good practice considering that in this case the constant is an object not a primitive. But sure, let's not care until there's a reason to care (seriously).

dmitrysteblyuk:
I restored pretty much the previous behavior. Now the changes address the required fix only.

sakari (approved):

PR #444, opened 2023-08-24
Support abort signal

xmahle:
**What?**
Allow passing `AbortController.signal` to oats axios / fetch adapters.

### DIFF
--- a/test/fetch-adapter/fetch-adapter.spec.ts
+++ b/test/fetch-adapter/fetch-adapter.spec.ts
@@ -7,12 +7,15 @@ import * as Koa from 'koa';
 import { koaBody } from 'koa-body';
 import * as fetchAdapter from '@smartlyio/oats-fetch-adapter';
 import * as http from 'http';
-import { setTimeout } from 'timers/promises';
 
 describe('fetch adapter', () => {
   let apiClient: client.ClientSpec;
   let httpServer: http.Server | undefined;
   let receivedContext: any;
+  let callback: (...args: any[]) => Promise<void>;
+  beforeEach(() => {
+    callback = async () => { return };
+  });
 
   const spec: server.Endpoints = Object.fromEntries(
     server.endpointHandlers.map(handler => [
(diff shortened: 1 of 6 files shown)


sakari:
seems sensible. could we have a test for this? Like in the adapters would be nice to check that the thing actually does what it promises and on oats side would be nice to see that the generated client passes the thing forward and does useful things with it.

xmahle:
> seems sensible. could we have a test for this? Like in the adapters would be nice to check that the thing actually does what it promises and on oats side would be nice to see that the generated client passes the thing forward and does useful things with it.

Yes!

sakari:
We could add some hook [here](https://github.com/smartlyio/oats/pull/444/files#diff-2badef9cc20a5025feecb6fdcbd5e87b282db660a177b02416b3e98f3fc5506fR22) to stop the request from resolving before signalling the abortcontroller. Otherwise this I think theoretically could be a bit flaky.

test/fetch-adapter/fetch-adapter.spec.ts
```diff
@@ -19,6 +20,7 @@ describe('fetch adapter', () => {
       {
         [handler.method]: async (ctx: any) => {
           receivedContext = ctx;
+          await setTimeout(100);
```
sakari, line 23: would 0245294db4de43f7b2baf8279d7a4abe21d780e4 avoid issues with using timers?

xmahle: That looks like the earlier version I had 🤔 ... Could we just get this PR merged with that commit or not, I don't like going back and forth and wasting everyones time. @sakari

sakari: you mean [this](https://github.com/smartlyio/oats/pull/444/commits/737db06501e65828966d448c11e11d76a2c0641b). As far as I can see with my commit we would not rely on timings whereas with that commit we would rely on the server not resolving the request before the abort controller fires.

sakari: fine with that. Could we avoid force pushing stuff while things are in flight because it makes collaboration way less fun.

xmahle: Yeah, bad habbit

sakari (approved):

PR #457, opened 2024-03-07
allow multiple non-objects with explode true

sakari:
see changelog

canary at 7.4.4-alpha.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,16 @@
-#
+# Next
+
+Allow expand=true with multiple query parameters when the query parameter schemas are not 
+
+ * references
+ * objects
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
valentin-nemcev, line 1: wait, you're supposed to fill changelog manually

valentin-nemcev (approved):

roginvs (changes requested):
[tested locally]

I would expect to see a unit test for this case, especially keeping in mind that this project already have good tests suite

CHANGELOG.md
sakari: yes. automating this would require more care in commits which I don't feel is good way to spend time

roginvs (approved):

PR #459, opened 2024-03-12
Support relative servers in fetch adapter in the browser

valentin-nemcev:
Currently, every fetch-based browser client prepends the location.origin to the already-specified relative path in the server definition, it's pure boilerplate. E.g. you can delete the `servers` prop from 
```
adapter.create()({
    ...spec,
    headers,
    servers: [`https://${window.location.host}/mosby`],
  })
```  
(people who put `http://localhost:9000/goya` into their servers field will continue enjoying their servers array overrides)

With this PR, fetch adapter will do the reasonable thing and use page origin if the no origin is provided. I.e. it modifies behavior only for a case when an error was thrown before, it will not override explicitly provided origins

No tests, as I'd be testing jsdom setup and built-in classes mostly, and I didn't want to add a whole new test adapter just for this.

### DIFF
--- a/packages/oats-fetch-adapter/index.ts
+++ b/packages/oats-fetch-adapter/index.ts
@@ -62,7 +62,7 @@ export function create(init?: RequestInit): runtime.client.ClientAdapter {
     }
     const server = arg.servers[0];
     const data = toRequestData(arg.body);
-    const url = new URL(server + arg.path);
+    const url = new URL(server + arg.path, globalThis.location?.origin);
     const requestContentType = arg.body?.contentType;
 
     Object.entries(arg.query ?? {})


smartly-petri (approved):

By-Ilya (approved):

valentin-nemcev:
> > you can delete the servers prop from
> 
> isn't that only overriding the servers value coming from the openapi3 specification. So if we do not override we will use the servers setting from there. What am I missing?

And the servers from there defined either like `http://localhost:9000/goya` which is technically a server, but is useless, or like `/mosby` which is technically not a server, but is actually useful, this PR adds support for the `/mosby` style declaration

I don't really understand how the servers array is supposed to be used, given that openapi files don't know where they are actually going to be deployed, only the path prefix, but at least it supports existing usage in a less non-sensical way

sakari:
they are supposed to be used as described [here](https://learn.openapis.org/specification/servers.html#the-server-object). I don't think a relative path is allowed there. We override those servers arrays to make things work in our testing environments  where the service urls differ

valentin-nemcev:
Ok, so we don't use the server thing as intended, but my addition doesn't say anything about servers, technically, it just lets you omit origin in the request url in browser context, IMO a reasonable behavior for a fetch adapter. @sakari do you think it's a useful addition? If not, let's just close it then, it's not a big deal at all

sakari:
kinda would err on the side of less magic given that this is not currently hard to diy like you show. I do see the less boilerplate benefit but we could provide similar thing as a helper outside oats in some frontend repository we have.

so close?

PR #460, opened 2024-04-05
Fix `data.getHeaders is not a function` issue

syedtauhidahmed:
Screenshot 2024-04-05 at 2 02 36 PM

[form-data](https://www.npmjs.com/package/form-data) is used only on Node. if its run on the browser, it will switch to the window's version of FormData. Library code:
`module.exports = typeof self == 'object' ? self.FormData : window.FormData`

For that reason we are getting the above error while making a `multipart/form-data` type request.

### DIFF
--- a/packages/oats-axios-adapter/index.ts
+++ b/packages/oats-axios-adapter/index.ts
@@ -92,7 +92,7 @@ function createAxiosAdapter({
     const url = server + arg.path;
     const headers = removeUndefined({
       ...arg.headers,
-      ...(data instanceof FormData ? data.getHeaders() : {})
+      ...(data instanceof FormData && data.getHeaders ? data.getHeaders() : {})
     });
     const response = await axiosInstance.request({
       method: arg.method,


packages/oats-axios-adapter/index.ts
```diff
@@ -92,7 +92,7 @@ function createAxiosAdapter({
     const url = server + arg.path;
     const headers = removeUndefined({
       ...arg.headers,
-      ...(data instanceof FormData ? data.getHeaders() : {})
+      ...(data instanceof FormData && data.getHeaders ? data.getHeaders() : {})
```
afarriaga, line 95: 👏

afarriaga (approved):

packages/oats-axios-adapter/index.ts
sakari: how do we set the correct content-type if we do not do it here?

syedtauhidahmed: When form-data is used as payload, its Content-Type defaults to `multipart/form-data` in browser. In the console, i executed this line to check out the Request header. 
`FormData.prototype.getHeaders = function() {return {}}`
image

sakari (approved):

PR #466, opened 2024-06-26
make endpoints type method handlers optional

dmitrysteblyuk:


sakari (commented):
will omitting the method handler work?

dmitrysteblyuk:
>will omitting the method handler work?

Yes, because we can specify this method in another router. Basically, here's the issue:

https://smartlyio.atlassian.net/browse/ENM-2068

sakari:
could you create a test for this following the example [here](https://github.com/smartlyio/oats/tree/master/test/koa-adapter)

dmitrysteblyuk:
> could you create a test for this following the example

Done 👍

sakari (approved):

PR #473, opened 2025-01-14
Upgrade to Node 20.18.1

xmahle:
Let's update Node version to more 'today' ... shall we? :D


sakari:
we could test against node 20 also as I think we are using that (well and 18 probably somewhere still 😁 )

xmahle:
> we could test against node 20 also as I think we are using that (well and 18 probably somewhere still 😁 )

Yeah, I just looked at the Node version chart, and based on that we should be using 22... But I'll change the PR to use 20.x first..

xmahle:
Node js — Node js Releases 2025-01-14 17-56-32

sakari:
@xmahle any idea why github says 
image

xmahle:
> @xmahle any idea why github says image

Good question, didn't find any reason with quick search..

xmahle:
Ah, oats-runtime has it's own `.node-modules` file... why is that 🤔

sakari (approved):
lets see if this is just github thing

PR #477, opened 2025-03-27
includeEndpoints-driver-option

dmitrysteblyuk:
Implements `includeEndpoints` driver option to generate code only for the given endpoints.

### DIFF
--- a/packages/oats/src/util.ts
+++ b/packages/oats/src/util.ts
@@ -1,8 +1,34 @@
 import * as oas from 'openapi3-ts';
 import * as assert from 'assert';
 import * as _ from 'lodash';
+import type { IncludeEndpointFilter } from './generate-server';
+import type { Driver } from './driver';
 
 export type SchemaObject = oas.ReferenceObject | oas.SchemaObject;
+
+export function createIncludeEndpointFilter(
+  includeEndpoints: Driver['includeEndpoints'],
+  spec: oas.OpenAPIObject
+): IncludeEndpointFilter {
+  if (includeEndpoints === undefined) {
+    return () => true;
+  }
+  const includeEndpointsMap = new Map(
+    Object.entries(includeEndpoints).map(([path, includeMethods]) => {
+      const methods = includeMethods.map(method => method.toLowerCase());
+
+      for (const method of methods) {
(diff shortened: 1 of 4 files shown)


sakari:
did you check whether we get good results with tree shaking using this option?

packages/oats/src/util.ts
```diff
@@ -1,9 +1,48 @@
...
 import * as assert from 'assert';
 import * as _ from 'lodash';
+import type { Driver } from './driver';
+import { server } from '@smartlyio/oats-runtime';
 
 export type SchemaObject = oas.ReferenceObject | oas.SchemaObject;
 
+export function filterEndpointsInSpec(
+  spec: oas.OpenAPIObject,
+  { includeEndpoints }: Pick<Driver, 'includeEndpoints'>
```
sakari, line 11: allowing this to be a function would allow a bit more ergonomics in filtering

sakari:
mmm. would we not get the same effect with something like [this](http://stackoverflow.com/questions/77096165/filtering-an-openapi-spec-using-jq)

packages/oats/src/util.ts
dmitrysteblyuk: we can support an additional function type, for example:
```typescript
export interface Driver {
  includeEndpoints?:
    | Record<string, Uppercase<serverRuntime.Methods>[]>
    | (path: string, method: Uppercase<serverRuntime.Methods>) => boolean;
}
```
if/when there's a need. For now, it's quite handy to have this as a simple record:
first, it's declarative
second, if you by mistake included a non-existing endpoint, you will get an error

sakari:
> i tried, but frontend is not the best place to test this.

soo. what is the benefit of doing this filtering then?

sakari (approved):

packages/oats/src/util.ts
dmitrysteblyuk: anyways, i did some refactoring like you suggested. So the function type is supported (not that I see use case for it so far).

sakari (approved):

dmitrysteblyuk:
@sakari how should i proceed? Simply merge or write changelog, bump the version etc?

PR #484, opened 2025-08-28
Update TypeScript to 5.9.2

lautis:
Update project typescript version to latest. This should fix issues when installing newer types.

I've cleaned up tsconfig from extra libs and type roots.


packages/oats-runtime/test/shape-of.spec.ts
```diff
@@ -82,9 +82,13 @@ describe('ShapeOf', () => {
   });
 
   it('works with nested branded types', () => {
+    // @ts-expect-error -- does not work with nested branded types
```
sakari, line 85: erm. so updating ts breaks ShapeOf? any idea why

lautis: No idea. This works however:


```
type NestedValue = runtime.BrandedScalar<'a' | 'b' | 'c' | 'd', BrandOfNestedValue>;
```

lautis: The update doesn't change the produced output at least for me. So, this PR is a no-op in that sense and only makes it easier to maybe address this and other issues.

PR #515, opened 2026-05-28
add oats playwright adapter

bezrodnov:
Add a new OATS adapter for playwright tests


sakari (approved):

bezrodnov:
I've just squashed all commits into one and force pushed changes cause I noticed that squash and merge option is not enabled for this repository and I don't have rights to change it.

PR #517, opened 2026-06-12
fix: support relative path in playwright adapter

bezrodnov:
Currently playwright adapter for oats throws an error if the adapter is configured using relative url (e.g. `/creative-templates` or `/matrix`). In this case it throws an [error](https://github.com/smartlyio/frontend/actions/runs/27412288256/job/81024282009?pr=51076#step:10:720): `TypeError: Invalid URL`.

This PR should fix this problem by using a similar approach as in the axios adapter.

The change was tested locally in frontend repository by manually changing compiled file to use the same code as in the changes ts file.

### DIFF
--- a/packages/oats-playwright-adapter/index.ts
+++ b/packages/oats-playwright-adapter/index.ts
@@ -159,19 +159,22 @@ export function create(
     }
     const server = arg.servers[0];
     const requestBody = toRequestBody(arg.body);
-    const url = new URL(server + arg.path);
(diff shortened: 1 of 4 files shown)


sakari:
is there a way to create a test case for this?

bezrodnov:
@sakari I've added some unit tests to cover this change. Could you please have a look?

sakari (approved):

PR #518, opened 2026-06-16
chore: add publish-only job

bezrodnov:
## Summary

Adds a manual recovery path to the Publish workflow for when the version commit and git tag are already on `master`, but npm publish failed or was only partially completed.

## Context

The v7.8.1 release hit this scenario: the version bump commit and `v7.8.1` tag were pushed successfully, but npm publish failed due to an invalid `NPM_TOKEN`. Re-running the failed workflow made things worse — GitHub Actions re-runs from the original merge commit, not current `master`, so the workflow tried to create the same version commit again and the push was rejected with a non-fast-forward error.

## Changes

- **`.github/workflows/publish.yml`**
  - Add `workflow_dispatch` trigger for manual recovery runs
  - Reuse the existing `publish` job with a `PUBLISH_ONLY` flag (`true` on `workflow_dispatch`, `false` on `push`)
  - When `PUBLISH_ONLY` is set, skip version bump, commit, tag, and GitHub App/git setup steps
  - Checkout current `master` directly and run the same build + publish steps as a normal release
- **`README.md`**
  - Document the partial publish failure corner case
  - Explain why re-running the failed workflow does not work
  - Document the recovery steps via **Actions → Publish → Run workflow**

## Test plan

Testing this change is not that straightforward. One way to do it is to delete npm token from registry so that publish step fails. And after that generate a new token and run publish workflow manually from github.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,13 +1,17 @@
 # Publish workflow — automated versioning and npm publishing.
 #
 # Contributors only need to add a label (patch/minor/major) to their PR.
-# After merge, this workflow:
+# After merge, the `publish` job:
 #   1. Determines the version bump type from merged PR labels
 #   2. Bumps all package versions in fixed mode
 #   3. Commits the version bump directly to master
 #   4. Publishes each package to npm (see .github/scripts/publish-packages.sh)
- #   5. Creates and pushes a git tag
- #
+#   5. Creates and pushes a git tag
+#
+# If publish fails after the version commit is already on master, use
+# "Run workflow" (workflow_dispatch) to re-run the same job with PUBLISH_ONLY
+# set — version bump, commit, and tag steps are skipped.
+#
 # Publishing uses `npm publish` directly (via a separate script) rather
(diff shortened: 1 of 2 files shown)


htrungngx:
cross-link PR to permanently fix the expired npm token issue by using npm trust publisher (OIDC) https://github.com/smartlyio/oats/pull/519

bezrodnov:
@htrungngx thank you for addressing the issue and removing the painful token rotation process 🙏 
As far as I can see the PR you merged to address it did not trigger a NPM publishing. And my previous PR with a fix that I want to get published to NPM registry failed publishing workflow. Could you please recommend a safe way to deploy my fix? Is there an approach that won't require a dummy PR?

htrungngx:
where is your PR ?

bezrodnov:
https://github.com/smartlyio/oats/pull/517

bezrodnov:
Hi @htrungngx any update? Should I consider re-opening this PR?

htrungngx:
let's do a new patch version bump, that's easier

